PR #19, opened 2026-05-17
ci: use 16-core GitHub runners

rabble:
## Summary
- move Ubuntu GitHub Actions jobs from `ubuntu-latest` to the org hosted `ubuntu-24.04-16core` runner
- keep workflow logic unchanged; this only changes runner capacity

## Motivation
Reduce CI/build lead time across Divine repos. The org runner `ubuntu-24.04-16core` is configured as Ubuntu 24.04, 16 cores, 64 GB RAM, max 50 concurrent runners.

## Manual validation
- Parsed changed workflow YAML with Ruby
- Ran `git diff --check -- .github/workflows`

### DIFF
--- a/.github/workflows/publish-and-release.yml
+++ b/.github/workflows/publish-and-release.yml
@@ -34,7 +34,7 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-16core
     services:
       redis:
         image: redis:7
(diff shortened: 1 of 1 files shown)


rabble:
Closing as obsolete. This May runner-size bump is stale/blocked and no longer matches the current CI cost direction; newer work is reducing runner size where appropriate instead of moving these repos to 16-core runners.

PR #22, opened 2026-06-08
fix(push): verify payload shape to avoid duplicate client rendering

dcadenas:
## Summary

iOS users saw two banners for a single push. This change makes the push service send one notification shape, so one push produces one banner.

The fix drops `content_available` from iOS alert pushes. That flag was waking the app's background isolate to draw a second, local banner on top of the one iOS already shows from `aps.alert`.

## Motivation

Every notification carried `aps.alert` + `content_available` + `mutable_content` together (added in #17). On iOS the OS presents the `aps.alert` banner, and `content_available` separately wakes the app to render its own — a duplicate.

`content_available` is only needed for silent pushes. An alert push still reaches terminated iOS apps without it, so removing it costs no delivery reliability, and `mutable_content` stays for Notification Service Extension enrichment. The mobile team reached the same root cause in divine-mobile#4760 and recommended this exact server change.

## Related Issue

Closes #20. Mobile contract: divine-mobile#4760 (defense-in-depth client guard, already merged). Follow (kind 3) notifications remain unimplemented and are tracked separately in #21.

## Testing

Ran the full Rust suite locally; all green. Not run on a physical iOS device (push is not deliverable to the simulator).

- `cargo test` — 64 tests pass
- `cargo clippy --all-targets --all-features` — clean
- `cargo fmt --check` — clean
- TDD: flipped the alert-payload unit test to assert `content-available` is absent (watched it fail, then pass).

## Protocol / Ops Notes

iOS payload contract change. Alert pushes now omit `content-available` (still `aps.alert` + `mutable-content`, push-type `alert`, priority 10). Android (data-only) is unchanged. The live mobile guard already suppresses duplicates, so this is the durable server-side layer, not the sole defense.

No partner, customer, or brand names appear in the title, description, branch, or code. Confirmed.

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -52,9 +52,14 @@ The service watches for these event kinds and notifies the tagged recipient:
 | Mention | 1 | Note mentioning user (p-tag, no e-tag reference) |
 | Repost | 16 | Repost of user's note (p-tag) |
 
+> **Note:** Follow (kind 3) is defined but **not currently emitted** — the handler skips kind 3 because new-follow detection requires diffing contact-list state, which is not yet implemented. Likes, comments, mentions, and reposts are the types actually delivered today.
+
 ## FCM Payload Format
 
-The service sends **data-only** FCM messages (no `notification` field). This gives the client full control over how notifications are displayed.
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the APNS payload change and docs, verified focused tests, fmt, check, clippy, and confirmed CI is green. This removes the duplicate-banner cause without changing the silent/background path.

PR #23, opened 2026-06-09
fix(push): make notification payload targets authoritative for mobile routing and attribution

dcadenas:
## Summary

Push notifications could point people at the wrong video, and comments on videos sent no push at all.

This change makes the FCM payload carry the video's real target: the full `kind:pubkey:d-tag` coordinate and its parts, taken from the triggering event's signed `a`/`A` tag. The owner now comes from the actor's event, not from whoever is being notified. It also starts sending pushes for NIP-22 video comments (kind 1111), which the service ignored before. The docs now spell out which payload fields are safe to route on and which are display-only.

## Motivation

Mobile (divinevideo/divine-mobile#4727, divinevideo/divine-mobile#4730) needs a payload it can route from without guessing who owns a video. Before, the payload named a d-tag but never the owner, so the client assumed the owner was the recipient. That is right only when you act on your own video, and wrong for a reply to someone else's comment. Video comments (kind 1111) were dropped entirely, which is exactly where wrong-owner guessing hurts most.

## Related Issue

Part of #21. The end-to-end fix also needs the mobile consumer (divinevideo/divine-mobile#5003), the funnelcake producer (divinevideo/divine-funnelcake#458), and kind-3 follow pushes (divinevideo/divine-funnelcake#432).

## Testing

`cargo test` (72 pass), `cargo clippy --all-targets --all-features` (clean), `cargo fmt --check`. Added unit tests for addressable-coordinate extraction (like/repost, NIP-22 root-scope preference, no-coordinate fallback, colons in d-tag) and for kind 1111 recipients (reply notifies root + parent authors, top-level dedup, no author tags). Verified real event shapes against relay.divine.video.

## Protocol / Ops Notes

Additive and non-breaking. New `data` keys are extra strings existing clients ignore: `referencedAddress`, `referencedKind`, `referencedAuthorPubkey`, `referencedDTag`; `referencedEventId` now prefers the NIP-22 root scope (`E`) over the parent `e`. The service now also subscribes to kind 1111 and notifies both the video owner (root `P`) and the parent comment author (parent `p`), deduplicated.

No partner, customer, or brand names or other sensitive identities appear in the title, branch name, or description.

### DIFF
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -37,6 +37,7 @@ notification:
     - 3      # Contact list (follows)
     - 7      # Reactions/likes (NIP-25)
     - 16     # Generic reposts (NIP-18)
+    - 1111   # NIP-22 comments
     - 30023  # Long-form content (mentions)
   # Default kinds users receive notifications for (can be customized via kind 3083)
   default_preferences:
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -38,6 +38,7 @@ notification:
     - 3      # Contact list (follows)
     - 7      # Reactions/likes (NIP-25)
     - 16     # Generic reposts (NIP-18)
+    - 1111   # NIP-22 comments
     - 30023  # Long-form content (mentions)
   # Default kinds users receive notifications for (can be customized via kind 3083)
   default_preferences:
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -48,11 +48,14 @@ The service watches for these event kinds and notifies the tagged recipient:
 |------|-----------|---------|
 | Like | 7 | Reaction to user's note (p-tag) |
 | Comment | 1 | Reply to user's note (p-tag, with e-tag reference) |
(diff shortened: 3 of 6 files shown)


NotThatKindOfDrLiz:
Pushed a small follow-up cleanup in c40103b after the final review pass.

What changed:
- The live Nostr subscription now uses `notification.event_kinds` from settings for notification-trigger kinds instead of keeping a separate hardcoded list in `nostr_listener.rs`.
- Control kinds remain explicit and are deduped with configured notification kinds.
- Added kind `1111` to the default/configured notification subscription kinds so NIP-22 comment handling is represented in config as well as code.
- Added a guard that skips out-of-range configured kinds with a warning.

Validation:
- `cargo fmt --check`
- `cargo check`
- GitHub checks are green: semantic PR title, CLA, and test.

NotThatKindOfDrLiz (approved):
Final pass is clean. The payload changes are additive/backward-compatible, kind 1111 recipient handling matches NIP-22 root/parent author semantics, and the follow-up commit removes the stale hardcoded notification-kind list by driving live subscriptions from config. GitHub checks are green.

PR #29, opened 2026-07-19
feat(push): deliver FCM push for "Inspired by" video references (subscribe kind 34236)

dcadenas:
## Summary

People tagged on a new video did not receive a push notification because the service did not subscribe to or handle kind 34236.
This change sends the existing `mention` push and includes the video's own coordinate so the app can open the right video.

## Motivation

A kind 34236 video is the trigger, so it does not carry an `a` or `e` tag that points back to itself.
The service now builds deep-link fields from the video's event id, author, and `d` tag.
Successful delivery is recorded per video coordinate and recipient only after at least one FCM token succeeds, so a failed send can be retried by a later edit.

## Related Issue

Closes #28

## Testing

The full Rust suite passes, including Redis-backed coverage for failed sends, successful edits, and later edit suppression.

- `cargo fmt --check`
- `REDIS_URL=redis://127.0.0.1:16379 cargo test`

## Protocol / Ops Notes

Production and development relay subscriptions now include kind 34236. Video p-tags emit type `mention` and remain governed by the existing kind 1 mention preference.
Recipient matching is deliberately marker-blind, so every valid p-tag is eligible. Successful coordinate deliveries use `video_coordinate_dedup_ttl_secs`, which defaults to one year.

The public branch name and description contain no partner, customer, campaign, or other sensitive external identity.


NotThatKindOfDrLiz (approved):
Approved after re-review. No blocking findings; CI is green. I did not merge.

PR #34, opened 2026-07-29
docs: plan new-post ("bell") push notifications

### LINKED ISSUE #33: feat(push): new-post ("bell") notifications for followed creators
## Problem

Every notification this service emits today is *"someone acted on your content"* — like, comment, repost, or mention, all resolved from `p` tags on the trigger event. Users have asked for the inverse: **alert me when a creator I care about posts**.

Nothing in the current model supports per-creator targeting. Preferences are a flat list of event kinds; the service has no notion of which creators a given user cares about.

## Proposed approach

Subscription list lives on Nostr as a NIP-51 kind 30000 people list with a reserved `d=notify` tag, published by the mobile client (mirroring the existing `d=block` precedent). This service subscribes to those lists, maintains a `creator → [subscribers]` reverse index in Redis, and on each incoming kind 34236 video notifies every subscriber of that video's author — rate-limited to one push per (subscriber, creator) per hour.

Kind 34236 is already subscribed and `insert_video_reference_fields` already emits the exact routing payload needed, so **no new FCM fields**.

Full plan: `docs/plans/new-post-notifications.md` (see linked PR).

## Scope

Seven tasks, each independently verifiable:

1. Subscribe to `d=notify` people lists (separate `#d`-filtered subscription; historical processing mandatory)
2. Ingest list events into the Redis reverse index (atomic Lua diff-and-apply with a `created_at` ordering guard)
3. Resolve new-post recipients (refactor to a flat `Vec<NotificationTarget>`; Mention wins on dedup)
4. Preference gating (`NewPost.kind() == 34236`)
5. Per-creator hourly rate limit
6. FCM copy (provisional pending brand-guidelines sign-off)
7. Docs, including `AGENTS.md`

## Cross-repo dependencies

- **`divine-mobile`** — bell UI, `d=notify` list publishing, and including 34236 in its kind-3083 preferences event. **Hard ordering dependency:** a partial rollout that publishes notify lists without updating preferences silently gates every new-post push off, and it will look like a service bug.
- **`divine-funnelcake`** — in-app feed row for belled creators (unthrottled; the rate limit is push-only).

## Accepted tradeoff

The `d=notify` list is world-readable, so who you have belled is public. This was a product decision in favor of portability, not an oversight — encrypting the content field would make the list unreadable to this service and break the feature.

rabble:
## Summary

Adds `docs/plans/new-post-notifications.md`, an implementation plan for new-post ("bell") push notifications. **Docs only — no code, no behavior change.**

The plan breaks the feature into seven independently verifiable tasks, each with the files it touches, the specific pitfalls, and the tests it needs.

Closes #33 is *not* claimed here — this PR only lands the plan; the issue stays open to track implementation.

## Motivation

Every notification this service emits today is "someone acted on your content" — like, comment, repost, mention — all resolved from `p` tags on the trigger event. Users asked for the inverse: alert me when a creator I care about posts. That is a subscription to someone else's output, and nothing in the current model supports per-creator targeting.

Landing the plan as its own PR gets the design decisions reviewed before any code exists, and gives the `divine-mobile` and `divine-funnelcake` work a written contract to build against.

## Linked issue

Tracking issue: #33

## Approach in brief

- Subscription list is a NIP-51 kind 30000 people list with a reserved `d=notify` tag, published by the mobile client — mirroring the existing `d=block` precedent. Public and portable.
- This service subscribes with a `#d`-narrowed filter, maintains a `creator → [subscribers]` reverse index in Redis, and on each incoming kind 34236 video notifies that author's watchers.
- One push per (subscriber, creator) per hour.

Two things make this smaller than it looks, and both are already in `main`:

- Kind 34236 is already subscribed (landed in #29).
- `insert_video_reference_fields` already emits the exact routing payload the tap router needs — `referencedAddress`, `referencedEventId`, `referencedKind`, `referencedAuthorPubkey`, `referencedDTag`. **No new FCM fields.**

## Decisions worth reviewer attention

These are the parts where I'd most like a second opinion:

- **Public subscription lists.** `d=notify` is world-readable, so who you have belled is public. Accepted product tradeoff for portability, written into the Risks section so nobody later "fixes" it by encrypting the content field — that would make the list unreadable to this service and kill the feature.
- **Check-then-set rate limiting** rather than atomic `SET NX EX`. Mirrors the existing video-coordinate dedup, and means a failed FCM send does not burn the user's hour-long window. The cost is a rare double-send race across replicas. The plan requires this rationale in a code comment.
- **Non-cluster-safe Lua script.** The diff-and-apply writes `notify_watchers:*` keys not declared in `KEYS`. Fine on single-instance Redis (`docker-compose.yml`), and the plan requires a comment saying so.
- **No server-side backfill**, deliberately. Reaching the gated code path requires having belled someone, which requires the new mobile build, which publishes preferences and the notify list together.
- **Mention wins on dedup.** A user who both watches a creator and is mentioned in the video gets one push, typed `Mention`.

## Manual test plan

This PR is a Markdown file, so verification is review-level:

- [x] `git diff main...HEAD --stat` shows exactly one added file, `docs/plans/new-post-notifications.md` (379 insertions, 0 deletions)
- [x] No `src/`, `config/`, `tests/`, or CI paths touched — nothing to build or run
- [x] Every file:line reference in the plan checked against `main` (`preferences.rs:14`, `preferences.rs:50`, `event_handler.rs:192`, `:384`, `:420`, `:533`, `:624`, `:634`, `:732`, `:804`, `:937`, `config.rs:64`, `config/settings.yaml:44`)
- [ ] Reviewer confirms the four Redis key families don't collide with existing `divine:*` keys
- [ ] Reviewer sanity-checks the `Filter::identifier` claim against nostr-sdk 0.44

The end-to-end test plan for the *implementation* lives in the plan's Verification section: bell a creator, assert the reverse index, post a video, confirm the push and its tap target, post a second video and confirm the rate-limit skip, unbell and confirm delivery stops.

## Rollout notes

Ship order is safe in both directions on its own — service-first finds no watchers, client-first starts delivering when the service deploys. The one unsafe case is a *partial* mobile rollout that publishes notify lists without also including 34236 in its kind-3083 preferences: new-post pushes are silently gated off and it looks like a service failure. Called out in both the plan and #33.

Companion work: `divine-mobile` (bell UI + list publishing), `divine-funnelcake` (in-app feed row).

### DIFF
--- a/docs/plans/new-post-notifications.md
+++ b/docs/plans/new-post-notifications.md
@@ -250,11 +250,25 @@ with `display_name() == "newPost"` and `kind() == 34236`.
 
 The `display_name` string is a wire contract — `divine-mobile` matches on
 it in `notificationKindFromPushType`
-(`mobile/lib/notifications/routing/notification_tap_target.dart`), which
-returns `null` for unrecognized values and degrades the tap target. It
-does not yet know `newPost`. `parseFcmPayload` is type-agnostic, so the
-push still displays; only the tap route is wrong. Do not change this
-string casually.
+(`mobile/lib/notifications/routing/notification_tap_target.dart:90`),
+which recognizes exactly five lowercase values
+(`like`/`comment`/`follow`/`mention`/`repost`) and returns `null` for
+anything else.
+
+An unrecognized value is **non-fatal, and the tap still routes
+correctly.** `resolveNotificationTapTarget` sends any non-`follow`,
+non-`system` kind with a video target to `OpenVideoTarget`, and `null` is
+neither (`notification_tap_target.dart:145-166`). Because this service
+always emits `referencedAddress` for a kind-34236 trigger,
+`hasVideoTarget` is true, so the notification opens the video with
+`autoOpenComments` false — correct for a new post. No client change is
+needed for routing to work.
+
+What the string does drive is the `NotificationKind` mobile uses for
+in-app row typing, so still do not change it casually. Note every value
+mobile recognizes today is lowercase, which argues for `newpost` over
+`newPost`; that is a cross-repo contract decision, not a detail to settle
+by whichever repo lands first.
 
 **Structural change required.** `handle_content_event` currently
 computes a single `(NotificationType, Vec<PublicKey>)` tuple
--- a/docs/plans/new-post-notifications.md
+++ b/docs/plans/new-post-notifications.md
@@ -153,7 +153,18 @@ if kind_num == KIND_NOTIFY_LIST {
 
 Add `redis_store::replace_notify_subscriptions(pool, subscriber, creators, created_at)`.
 The diff-and-apply must be atomic across replicas — do it in a **Lua
-script** keyed on the subscriber:
+script** keyed on the subscriber.
+
+Why atomic, concretely: production runs **2 replicas**
+(`divine-iac-coreconfig`, `k8s/applications/divine-push-service/overlays/production/kustomization.yaml`;
+staging and poc run 1, no HPA). `try_claim_event` only prevents two
+replicas handling the *same* event — two different list events from one
+subscriber can still land concurrently, and a read-then-write lets the
+older one win, resurrecting an unbell. `MULTI`/`EXEC` cannot express
+this: it queues blind writes, so the old set would have to be read
+outside the transaction. `WATCH` + retry would also work but is more
+code over pooled connections. A single-replica deployment needs none of
+this — the handler loop is sequential within a process.
 
 ```
 -- KEYS[1] = notify_subs:{sub}, KEYS[2] = notify_subs_ts:{sub}
@@ -170,13 +181,28 @@ subscriber from `notify_watchers:{removed}`, `SADD`s to
 (`docker-compose.yml`), so that is acceptable — **add a comment saying
 so** and gate on it if the deployment ever moves to Redis Cluster.
 
+**Bound the creator list before it reaches Redis.** The script runs as
+one blocking unit and Redis is single-threaded, so a list with thousands
+of `p` tags stalls the instance for every user. Add config
+`notify_list_max_creators` (default `1000`, well above any follow-gated
+bell list) and truncate with a warning rather than rejecting the list —
+the user keeps the bells that fit instead of losing all of them, and `p`
+tags are client-ordered so which survive is deterministic. This matters
+more than the cluster-safety note above: Cluster is hypothetical, an
+oversized list is something a user can do today.
+
 An empty `p` tag list is legitimate (user unbelled everyone) and must
 clear the forward set and remove them from every reverse index. Do not
 treat empty as "malformed, skip".
 
 **Tests:** `d`-tag rejection; self-reference dropped; add/remove diff
 produces the right reverse-index membership; an older `created_at` is
-ignored; empty list clears everything.
+ignored; empty list clears everything; the creator list truncates at the
+cap and duplicates do not consume cap budget.
+
+Keep the collection logic (dedup, self-filter, cap) in a **pure sync
+function** rather than inline in the async handler, or none of it is
+testable without a live Redis and an `AppState`.
 
 ### 3. Resolve new-post recipients
 
--- a/docs/plans/new-post-notifications.md
+++ b/docs/plans/new-post-notifications.md
@@ -185,9 +185,13 @@ ignored; empty list clears everything.
 Add `NotificationType::NewPost` to the enum (`preferences.rs:50`),
 with `display_name() == "newPost"` and `kind() == 34236`.
 
-The `display_name` string is a wire contract — `divine-mobile` matches
-on it in `parseFcmPayload` (`mobile/lib/services/notification_helpers.dart`).
-Do not change it casually.
+The `display_name` string is a wire contract — `divine-mobile` matches on
+it in `notificationKindFromPushType`
+(`mobile/lib/notifications/routing/notification_tap_target.dart`), which
+returns `null` for unrecognized values and degrades the tap target. It
+does not yet know `newPost`. `parseFcmPayload` is type-agnostic, so the
+push still displays; only the tap route is wrong. Do not change this
+string casually.
 
 **Structural change required.** `handle_content_event` currently
 computes a single `(NotificationType, Vec<PublicKey>)` tuple
@@ -239,14 +243,25 @@ Add `34236` to `UserPreferences::default()` and to
 **No server-side backfill is needed, and none should be written.**
 Users with a stale `user_preferences:*` entry lacking 34236 would be
 silently gated off — but reaching this code path at all requires having
-belled someone, which requires the new mobile build, which publishes
-kind 3083 including 34236 at the same moment it publishes the notify
-list. The client-side ordering makes the migration unnecessary.
+belled someone, which requires the new mobile build. Provided that build
+publishes kind 3083 including 34236 alongside the notify list, the
+client-side ordering makes a migration unnecessary.
 
-**This is a hard dependency on the mobile side.** Confirm the mobile PR
-publishes preferences alongside the first bell before shipping this
-service change, or new-post pushes will be gated off for every existing
-user and the bug will look like a service failure.
+**This is a hard dependency on mobile, and as of 2026-07-29 mobile
+cannot satisfy it.** `NotificationPreferences`
+(`mobile/lib/models/notification_preferences.dart`) is five fixed
+booleans, and `toKindsList()` can only emit a subset of `{1, 3, 7, 16}`.
+`push_notification_service.dart:254` publishes exactly that list, so
+**34236 is never published today** and this check gates every new-post
+push off.
+
+That is not a rollout-timing risk to be confirmed before shipping — it
+is the current state, and it fails silently. Mobile needs a sixth
+preference flag mapping to 34236 before any bell can produce a push.
+Tracked in the mobile design doc
+(`divine-mobile/docs/plans/2026-07-29-bell-notifications-design.md`,
+finding 2). Shipping this service change first is still safe: it finds
+no watchers and does nothing.
 
 **Tests:** `NewPost.kind() == 34236`; `NewPost.is_enabled()` false when
 prefs omit 34236; enabled under the shipped defaults.
@@ -373,7 +388,18 @@ the content field — that would make the list unreadable to this service
 and break the feature entirely.
 
 **Client dependency.** Task 4 is gated on the mobile client publishing
-34236 in its preferences. Ship order matters: service first is safe (it
-simply finds no watchers), client-first is safe (pushes start when the
-service deploys). Only a *partial* client rollout that publishes notify
-lists without updating preferences produces silent failure.
+34236 in its preferences, which it does not do today (see task 4). Ship
+order between the repos is still safe in both directions: service first
+finds no watchers and does nothing, client-first starts delivering when
+the service deploys. The unsafe case is a *partial* client rollout that
+publishes notify lists without the preference change — pushes are gated
+off silently and it looks like a service failure. That is the current
+mobile state, so the mobile release must carry both.
+
+**Reserved `d` tag collides with the mobile list UI.** `d=notify` decodes
+as an ordinary user-editable people list in `divine-mobile`
+(`Nip51PeopleListCodec` excludes only `d=block`), so before the mobile
+fix lands a user can rename or delete the list and silently wipe their
+own subscriptions. Nothing the service can defend against — it just sees
+a replacement list — but it is why the service must treat an empty `p`
+list as legitimate (task 2) rather than assuming malformed input.
--- a/docs/plans/new-post-notifications.md
+++ b/docs/plans/new-post-notifications.md
@@ -114,10 +114,8 @@ events when `context == EventContext::Historical`
 (`event_handler.rs:192`), but notify lists must be rebuilt from history
 or the reverse index cannot be recovered after Redis data loss — every
 subscription stays dark until each user happens to republish. (A process
-restart on its own is fine: Redis is external, so the index survives
-it. Defending the requirement on its real grounds makes it harder to
-argue away later.) Treat them like control events: process in both
-paths.
+restart on its own is fine — Redis is external, so the index survives
+it.) Treat them like control events: process in both paths.
 
 Make the historical lookback for notify lists independent of
 `process_window_days` (7 days). A list published 3 months ago and never
--- a/docs/plans/new-post-notifications.md
+++ b/docs/plans/new-post-notifications.md
@@ -112,8 +112,12 @@ extra filter; `subscribe` accepts one filter per call, so issue a second
 **Historical processing is mandatory here.** `route_event` skips content
 events when `context == EventContext::Historical`
 (`event_handler.rs:192`), but notify lists must be rebuilt from history
-or a restart silently drops every subscription until each user happens
-to republish. Treat them like control events: process in both paths.
+or the reverse index cannot be recovered after Redis data loss — every
+subscription stays dark until each user happens to republish. (A process
+restart on its own is fine: Redis is external, so the index survives
+it. Defending the requirement on its real grounds makes it harder to
+argue away later.) Treat them like control events: process in both
+paths.
 
 Make the historical lookback for notify lists independent of
 `process_window_days` (7 days). A list published 3 months ago and never
@@ -121,6 +125,23 @@ touched since is still current — it is a replaceable event. Use no
 `since` bound at all for the historical notify-list query, and add
 config `notify_list_history_limit` (default `5000`) as a safety valve on
 result size.
+
+**Dropping `since` is not sufficient, and this is still unresolved.**
+`run()` calls `is_event_too_old(&event)` on *every* event before it
+reaches `route_event` (`event_handler.rs:93`), against a hard-coded
+`REPLAY_HORIZON_DAYS = 7` (`event_handler.rs:40`). A 90-day-old
+`d=notify` event is dropped in the handler loop whatever the filter says
+and whatever order `route_event` uses, so as written above the rebuild
+recovers nothing older than a week — exactly the case it exists to
+protect.
+
+The mechanism is the author's call: a kind-aware horizon, an explicit
+exemption for list events, or something else. Whichever way it goes,
+task 1 needs a boundary test either side of the horizon. Note also that
+`try_claim_event` (`event_handler.rs:105`) runs before `route_event` and
+writes `dedup:{event_id}` with a 7-day TTL, so replaying a list event
+that was already processed is dropped as a duplicate — the plan should
+say how the rebuild is meant to interact with that.
 
 ### 2. Ingest list events into the reverse index
 
--- a/docs/plans/new-post-notifications.md
+++ b/docs/plans/new-post-notifications.md
@@ -196,23 +196,19 @@ The script reads the old set, computes added/removed, `SREM`s the
 subscriber from `notify_watchers:{removed}`, `SADD`s to
 `notify_watchers:{added}`, replaces `notify_subs`, and sets
 `notify_subs_ts`. Writing to `notify_watchers:*` keys not declared in
-`KEYS` is not Redis-Cluster-safe. That is acceptable here, but for a
-production reason rather than the local `docker-compose.yml`: the
-deployment points at
-`redis://redis-replication-master.redis-clusters.svc.cluster.local:6379/2`
-(`divine-iac-coreconfig`,
-`k8s/applications/divine-push-service/base/deployment.yaml:44`), and
-`k8s/redis-clusters/base/cluster.yaml` declares a `RedisReplication`
-plus a `RedisSentinel` — master/replica with Sentinel failover, one
-keyspace, no sharding. Cross-key writes are fine. **Add a comment saying
-so**, and say it in those terms: the namespace is called
-`redis-clusters` but is not Redis Cluster, which is exactly the sort of
-thing a future reader gets wrong in the unsafe direction. Gate on it if
-the deployment ever moves to real Cluster mode.
+`KEYS` is not Redis-Cluster-safe. That is acceptable here, but justify it
+from the production deployment rather than the local
+`docker-compose.yml`: production Redis is a **master/replica deployment
+with Sentinel failover — a single keyspace, not a sharded Cluster** (see
+the platform IaC repo for the manifests). Cross-key writes are fine.
+**Add a comment saying so**, and word it so the distinction survives:
+the manifests live under a *cluster*-flavoured name while not being
+Redis Cluster, which is what a future reader gets wrong in the unsafe
+direction. Gate on it if the deployment ever moves to real Cluster mode.
 
-That Redis is also **shared** — this service uses db index 2 of an
-instance other services key off, and Sentinel is watching its liveness.
-Anything that blocks it blocks them.
+That Redis is also **shared with other services**, on a dedicated
+logical database rather than a dedicated instance, with Sentinel
+watching its liveness. Anything that blocks it blocks them.
 
 **Bound the creator list before it reaches Redis.** The script runs as
 one blocking unit and Redis is single-threaded, so a list with thousands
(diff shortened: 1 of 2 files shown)


docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+events when `context == EventContext::Historical`
+(`event_handler.rs:192`), but notify lists must be rebuilt from history
+or a restart silently drops every subscription until each user happens
+to republish. Treat them like control events: process in both paths.
+
+Make the historical lookback for notify lists independent of
+`process_window_days` (7 days). A list published 3 months ago and never
+touched since is still current — it is a replaceable event. Use no
+`since` bound at all for the historical notify-list query, and add
+config `notify_list_history_limit` (default `5000`) as a safety valve on
```
dcadenas, line 124: Process every current `d=notify` list during rebuild. The `limit=5000` safety valve returns only the newest matching events, so older subscriptions are missing after bootstrap or index loss and will not receive pushes until republishing. Keep memory and relay load bounded with pagination, partitioning, or another durable recovery mechanism, and test a rebuild that crosses the batch boundary.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+
+`handle_notify_list_update`:
+
+1. Reject unless the `d` tag is exactly `"notify"` (defense in depth —
+   the relay filter should already guarantee this, but a malicious or
+   buggy relay can send anything).
+2. Extract `p` tag values, parse as `PublicKey`, dedup, drop
+   self-references (belling yourself is meaningless).
+3. **Replaceable ordering guard:** compare `event.created_at` against
+   `notify_subs_ts:{author}`. If the stored value is `>=` the incoming
```
dcadenas, line 167: Store the winning event id with `created_at` and apply NIP-01's lexicographically lowest-id tie-break inside the atomic update. Timestamps have only second resolution, so dropping every event with an equal timestamp makes the result depend on arrival order and can keep an unbell subscription active. Test both delivery orders and verify they produce the same forward and reverse indexes.

dcadenas (changes requested):
Complete the historical notify-list rebuild and make replaceable-event ordering deterministic before merging. The current 5,000-event cap drops older subscriptions during recovery, and timestamp-only ordering can ignore same-second unbell replacements. Add bounded pagination or an equivalent recovery strategy and the NIP-01 event-id tie-break, with tests covering both cases.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+there would wrongly apply to every kind in the list. Both
+`process_historical_events` and `subscribe_to_live_events` need the
+extra filter; `subscribe` accepts one filter per call, so issue a second
+`subscribe` call.
+
+**Historical processing is mandatory here.** `route_event` skips content
+events when `context == EventContext::Historical`
+(`event_handler.rs:192`), but notify lists must be rebuilt from history
+or a restart silently drops every subscription until each user happens
+to republish. Treat them like control events: process in both paths.
```
realmeylisdev: `route_event`'s historical skip isn't the only gate. `run()` calls `is_event_too_old(&event)` at `event_handler.rs:93` for **every** event before `route_event` runs, against a hard-coded `REPLAY_HORIZON_DAYS = 7` (`event_handler.rs:40`). Routing kind 30000 ahead of the control block doesn't help — the drop happens earlier, in the handler loop.

Confirmed locally: a kind-30000 `d=notify` event dated 90 days ago returns `is_event_too_old == true`; the same event dated 6 days ago returns false. So "use no `since` bound at all" recovers nothing older than a week — exactly the case the next paragraph exists to protect.

Task 1 needs an explicit horizon exemption for notify lists (kind-aware horizon, or a bypass for list events) plus a boundary test.

Two smaller notes on the same paragraph:

- `try_claim_event` (`event_handler.rs:105`) writes `dedup:{event_id}` with a 7-day TTL, so a historical replay of an already-processed list event is dropped as a duplicate. Worth saying how the rebuild is meant to interact with that.
- "a restart silently drops every subscription" isn't quite the failure mode — a restart preserves Redis, so the index survives it. The rebuild matters on Redis data loss. Defending the requirement on its real grounds makes it harder to argue away later.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+4. Post a second video as B within the hour; confirm no second push and
+   that the log records the rate-limit skip.
+5. Unbell B; confirm `notify_watchers:{B}` no longer contains A and no
+   further pushes arrive.
+
+## Risks
+
+**Relay volume.** The `#d=notify` filter keeps list ingestion cheap, but
+the service already receives every kind-34236 event on the relay. This
+change adds one `SMEMBERS` per video. Should that become hot, cache
```
realmeylisdev, line 462: "adds one `SMEMBERS` per video" undercounts this. The fan-out *after* the SMEMBERS is the expensive part, and it's serial end to end:

- `handle_content_event` iterates recipients one at a time, awaiting `send_notification_to_user` per recipient (`event_handler.rs:441-472`)
- each iteration is `get_tokens_for_pubkey` + `get_user_preferences` + the coordinate check + the FCM send
- `send_batch` loops tokens serially too (`fcm_sender.rs:286`)
- `run()` awaits `route_event` before pulling the next event off the mpsc(1000) channel (`event_handler.rs:126`)

Today the recipient list is the p-tags on an event — bounded and small. Bells make it every watcher of the creator, unbounded, so one popular creator posting stalls likes, comments and mentions for every other user until the fan-out drains.

This wants a decision in the plan rather than "do not pre-optimize": a config cap on watchers notified per video, a bounded-concurrency send, or an explicit statement of the fan-out size the design accepts.

Related: task 5 puts the rate-limit check alongside the coordinate check at `event_handler.rs:634`, which is after the token and preference reads. A rate-limited watcher costs three Redis round-trips to decide to do nothing — cheap to reorder while the code is being written.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+}
+```
+
+`handle_notify_list_update`:
+
+1. Reject unless the `d` tag is exactly `"notify"` (defense in depth —
+   the relay filter should already guarantee this, but a malicious or
+   buggy relay can send anything).
+2. Extract `p` tag values, parse as `PublicKey`, dedup, drop
+   self-references (belling yourself is meaningless).
```
realmeylisdev, line 165: Deliberately skipping `is_event_for_service` is the right call, but it leaves ingest with no authorization and no size bound. Any pubkey on the relay can publish a kind-30000 `d=notify` with N `p` tags, and the service writes N `SADD`s across N distinct `notify_watchers:*` keys. Repeat from fresh keypairs and Redis grows without limit on unauthenticated input — and none of those entries can ever produce a push, because the author has no registered token.

Worth adding to this step list: a cap on `p`-tag count (reject or truncate past it, with a log), and/or only indexing authors who already have a registered token — with a note on how someone who bells before registering gets picked up. Either is defensible; the plan should say which and why.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+
+| Key | Type | Purpose |
+|---|---|---|
+| `notify_subs:{subscriber_hex}` | SET | Creators this user has belled. Needed to diff against an incoming replacement list. |
+| `notify_subs_ts:{subscriber_hex}` | STRING | `created_at` of the last applied list event. Guards against out-of-order relay delivery of a replaceable event. |
+| `notify_watchers:{creator_hex}` | SET | Subscribers watching this creator. The hot read path — one `SMEMBERS` per incoming video. |
+| `notify_rate:{subscriber_hex}:{creator_hex}` | STRING w/ TTL | Rate-limit window marker. |
+
+`notify_subs` and `notify_watchers` are two views of the same relation;
+they must be updated together (see task 2).
```
realmeylisdev, line 77: Nothing in the plan ever deletes from these. `handle_deregistration` removes the token and deletes `user_preferences:*` (`event_handler.rs:308-311`), but a deregistered user stays in `notify_subs:*` and in every `notify_watchers:*` they appear in, indefinitely. `cleanup_service.rs` — the existing stale-key GC — isn't in any task's file list, and only `notify_rate:*` carries a TTL.

That's monotonic growth, and it keeps a record of who someone watched after they've left. Given the Risks section already treats this list as sensitive, the plan should either add deregistration cleanup as a task or state plainly that the index is never pruned and why that's acceptable.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+The rate limit is **push-only**. The in-app feed (FunnelCake) shows
+every post from belled creators. A user who receives one push for a
+six-post burst opens the app and sees all six — that is intended.
+
+**Interaction with existing video dedup, verify explicitly:** the
+per-recipient coordinate claim (`video_recipient_claim_key`,
+`event_handler.rs:533`) is keyed on
+`{kind}:{author}:{d_tag}:{recipient}` and already prevents a NIP-33
+*edit* of the same video from re-notifying. That protection applies to
+NewPost targets for free. Confirm with a test rather than assuming it.
```
realmeylisdev, line 377: "applies to NewPost targets for free" is true in the direction you're testing, and has a side effect in the other. `video_recipient_claim_key` is `{kind}:{author}:{d_tag}:{recipient}` (`event_handler.rs:533-541`) with no notification-type component, and it's set for a year (`video_coordinate_dedup_ttl_secs: 31536000`).

So: a watcher gets a NewPost push for video V, which burns `34236:B:<d>:A`. B later edits V to add a `p` tag mentioning A. That mention is suppressed at `event_handler.rs:634-663` — where a non-watcher would have received it. Belling a creator quietly costs you mention notifications on their edits.

Either scope the claim key by notification type, or make it a stated rule in the plan (with a test) that a watcher never gets a mention push for a video they've already been notified about. The listed tests only cover the suppression direction.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+
+### 3. Resolve new-post recipients
+
+**Files:** `src/event_handler.rs`, `src/preferences.rs`
+
+Add `NotificationType::NewPost` to the enum (`preferences.rs:50`),
+with `display_name() == "newPost"` and `kind() == 34236`.
+
+The `display_name` string is a wire contract — `divine-mobile` matches
+on it in `parseFcmPayload` (`mobile/lib/services/notification_helpers.dart`).
```
realmeylisdev: Two corrections. `parseFcmPayload` only surfaces the string — it doesn't match on it. The switch that matters is `notificationKindFromPushType` in `mobile/lib/notifications/routing/notification_tap_target.dart:90-107`, and its doc comment pins the contract as "a lowercase, five-value vocabulary (`like`/`comment`/`follow`/`mention`/`repost`)". `newPost` would be value six and the first camelCase one.

The good news, which the plan undersells: an unknown type is non-fatal there — it maps to `null`, and `resolveNotificationTapTarget` with a video target still returns `OpenVideoTarget`. Taps route correctly on the currently shipped client with no mobile change at all. Worth stating outright; it's a stronger claim than "do not add new payload fields".

Suggested edits: fix the file reference; add "extend the push-type vocabulary and its doc comment" to the client-responsibilities list at line 54; and if the lowercase convention is load-bearing, prefer `newpost` over `newPost`.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+reserved `d` tag, published by the mobile client:
+
+```
+kind 30000
+["d", "notify"]
+["title", "Notify"]
+["p", "<creator-pubkey-hex>"]   × N
+```
+
+This mirrors the existing `d=block` reserved-list precedent already used
```
realmeylisdev, line 31: The `d=block` precedent is on its way out. `content_blocklist_repository.dart` calls it "a non-standard kind 30000 list", carries `TODO(codex): Remove legacy kind 30000 publishing after kind 10000 rollout`, and already migrates users onto the standard kind 10000 mute list.

The kind-30000 choice here is fine on its own merits — NIP-51 defines 30000 as Follow sets keyed by an arbitrary `d`, which is exactly this shape. Cite NIP-51 rather than a precedent the mobile team is deleting; otherwise the justification evaporates the moment that TODO lands.

Separately: `d` values in kind 30000 are a shared, user-facing namespace — other clients let people name follow sets freely, so a user can already have a set called "notify" that has nothing to do with Divine. Step 1 of task 2 only checks `d == notify`, so that set would be read as a bell list. An app-scoped identifier would be safer.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+
+- Publishing and maintaining the `d=notify` list (add on bell, remove on
+  unbell, remove on unfollow — the bell is follow-gated).
+- Including kind `34236` in its kind-3083 preferences event. See
+  [Preference gating](#4-preference-gating) for why this matters and why
+  no server-side backfill is needed.
+
+## Redis schema
+
+Three new key families. All hex pubkeys, lowercase, full length — never
```
realmeylisdev: Four families in the table below (the PR body and line 328 both say four).

While here — the PR's test-plan item asks the reviewer to confirm no collision with existing `divine:*` keys. There are no `divine:*` keys. The real prefixes are `user_tokens:`, `user_preferences:`, `dedup:`, plus the fixed `stale_tokens` and `token_to_pubkey` (`redis_store.rs:21-27`, `preferences.rs:88`). No collision with `notify_*` either way, so the check passes — but it passes for a different reason than the checklist assumes.

The mismatch comes from `AGENTS.md`, which documents `divine:token:{pubkey}`, `divine:preferences:{pubkey}` and `divine:dedup:{...}` — none of which exist. Task 7 should correct those prefixes, not just append the new rows.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Verification
+
+```bash
+cargo clippy --all-targets --all-features
+cargo test
+```
+
+Redis-backed integration tests follow the existing `tests/dedup_test.rs`
+convention (skip cleanly when Redis is unavailable). Add
```
realmeylisdev: Worth naming which helper to copy. `dedup_test.rs` skips cleanly because it PINGs before handing back the pool. `preferences_test.rs` only checks that `create_pool` succeeded — which it does without a live server — and the tests then panic on the bb8 timeout. I hit exactly that locally before starting Redis: three failures, not three skips.

Pointing at `dedup_test.rs` is the right call. Adding "copy its PING-based helper, not the one in `preferences_test.rs`" would stop the new test file inheriting the broken pattern.

docs/plans/new-post-notifications.md
```diff
@@ -0,0 +1,379 @@
...
+Add `34236` to `UserPreferences::default()` and to
+`notification.default_preferences.kinds` in both
+`config/settings.yaml` and `config/settings.development.yaml`.
+
+**No server-side backfill is needed, and none should be written.**
+Users with a stale `user_preferences:*` entry lacking 34236 would be
+silently gated off — but reaching this code path at all requires having
+belled someone, which requires the new mobile build, which publishes
+kind 3083 including 34236 at the same moment it publishes the notify
+list. The client-side ordering makes the migration unnecessary.
```
realmeylisdev: `handle_preferences_update` overwrites wholesale — it parses the decrypted body into `UserPreferences` and calls `set_user_preferences` (`event_handler.rs:363-365`), no merge with what's stored.

So the client requirement is stronger than "publishes kind 3083 including 34236 at the same moment it publishes the notify list": **every** kind-3083 the client ever publishes afterwards must keep 34236, or new-post pushes silently switch off. Worth spelling out in the client-responsibilities list at line 54 — a later prefs write that drops the kind is precisely the silent failure this paragraph is trying to prevent.

realmeylisdev (changes requested):
Checked the plan against `main`, the nostr-sdk 0.44 source, NIP-51, and the `divine-mobile` client.

What holds up: all 13 file:line references resolve exactly. `Filter::identifier` exists (`nostr-0.44.2/src/filter.rs:679`) and maps to the `#d` custom tag, and `Client::subscribe` really does take one filter and mint a fresh subscription id per call — so the second-`subscribe` shape is right. "No new FCM fields" holds, and it holds harder than the plan claims: I traced a `newPost` payload through mobile's tap router and it routes to the video on the currently shipped build, no client change needed. `docker-compose.yml` is single-instance Redis, so the non-cluster-safe Lua caveat is accurate. Baseline is green locally with Redis 7 up — 79 tests, clippy clean.

Two gaps I'd want closed before this becomes implementation work.

**The 7-day replay horizon eats the historical rebuild.** Task 1 drops the `since` bound so a 3-month-old list still gets ingested, but `run()` calls `is_event_too_old` on every event at `event_handler.rs:93` — before `route_event`. Routing kind 30000 ahead of the control block doesn't get past it. I probed it locally: a 90-day-old `d=notify` event is rejected, a 6-day-old one isn't. This sits underneath @dcadenas's pagination point — even with perfect pagination, nothing older than a week reaches the handler.

**Fan-out is unbounded and serial.** `handle_content_event` walks recipients one at a time awaiting each send (`event_handler.rs:441`), `send_batch` loops tokens serially (`fcm_sender.rs:286`), and the handler awaits `route_event` before pulling the next event. Today "recipients" is the p-tags on an event — a handful. Bells make it every watcher of a creator, which head-of-line blocks likes and comments for everyone else. The Risks section calls this "one `SMEMBERS` per video"; the SMEMBERS is the cheap part.

Smaller ones inline: ingest has no size cap or authorization, nothing ever removes a user from the index, a NewPost push burns the coordinate claim a later Mention would have used, and the mobile wire-contract reference points at the wrong file.

Feedback only — no changes pushed to the branch.

realmeylisdev (changes requested):
Reviewer takeover. I pushed ten commits covering the findings whose remedy was a verified factual or internal-consistency correction, and left every finding whose remedy is a product or architecture decision with you. Docs only — nothing in `src/`, `config/`, `tests/`, or CI.

Gates checked first: org member with write access, in-repo branch rather than a fork, non-draft, mapping readable, changed paths inside the low-risk boundary. One commit per finding so you can revert any of them independently.

## Pushed

**`c28c1c6` — task 1's rebuild does not work as specified.** `run()` calls `is_event_too_old` on every event before `route_event` ever sees it (`event_handler.rs:93`), against a hard-coded `REPLAY_HORIZON_DAYS = 7` (`event_handler.rs:40`). Dropping the `since` bound therefore recovers nothing older than a week — the exact case the paragraph exists to protect — no matter how the filter or the routing order is written. I recorded the gap and the `dedup:` TTL interaction (`try_claim_event` runs before routing with a 7-day TTL) without picking a mechanism, and mirrored it into Prompt 1, which otherwise walks an implementer straight into the broken version. Also corrected the stated failure mode: a process restart preserves the index because Redis is external; the rebuild matters on Redis data loss.

**`2f4a0c6` — the creator cap never reached Prompt 2.** Design task 2 requires `notify_list_max_creators`, truncate-with-a-warning, the pure-sync collection helper and a truncation test. `31349d2` aligned the packet at 12:34; `6e50adf` added the cap at 12:40, so none of it propagated. Two documents in one PR disagreeing about a required step is worse than either being wrong alone, since the prompt is the copy-pasteable one.

**`b447563` — key-family count and AGENTS.md drift.** "Three new key families" over a four-row table. Separately, AGENTS.md's Redis Keys entries are all wrong today — `divine:` appears nowhere in `src/`; the real prefixes are `user_tokens:` and `dedup:` (`redis_store.rs:21-27`) and `user_preferences:` (`preferences.rs:88`), plus the fixed `stale_tokens` and `token_to_pubkey`. Task 7 now says to correct them rather than append four accurate rows to three inaccurate ones.

**`3e01d2d`, `b71f318`, `3304efa` — the Lua cluster-safety justification.** It argued production safety from `docker-compose.yml`, which is local dev. The conclusion holds for a different reason: production Redis is master/replica with Sentinel failover — one keyspace, not a sharded Cluster — so cross-key writes are fine. It is also shared with other services, which strengthens your own cap argument: an oversized blocking script stalls them too, not just this service's users. Same correction carried into Prompt 2.

**`54a40d7` — the tap-routing claim was inverted in both files.** Both said an unrecognized push `type` degrades the tap to a best-effort profile/inbox guess, and section 2.2 built a QA warning on it. An unknown value gives `kind == null`, which is neither `follow` nor `system`, so `resolveNotificationTapTarget` returns `OpenVideoTarget` whenever a video target is present (`notification_tap_target.dart:145-166`) — and this service always emits `referencedAddress` for a kind-34236 trigger. The tap opens the video either way, with `autoOpenComments` false, which is what a new post wants. So casing drives mobile's in-app row typing rather than routing, and end-to-end step 5 cannot catch a mismatch. I left the `newPost`/`newpost` choice open and flagged that every value mobile recognises today is lowercase.

**`173df84` — named the Redis test helper to copy.** `create_test_pool` (`dedup_test.rs:6`) issues a `PING` and returns `None`, so tests skip. `get_test_pool` (`preferences_test.rs:10`) only checks that `create_pool` returned `Ok`, which it does with nothing listening, and the tests then panic on the bb8 timeout. Pointing at `dedup_test.rs` was right but not specific enough to stop the new file inheriting the broken pattern.

**`07ec7e9`** — Prompt 4 cross-referenced a section title that does not exist ("Corrected:" vs "Superseded:"). **`d508867`** — prose trim.

## Left with you

These need a decision, not a reviewer's edit:

- **Both of @dcadenas's blocking items are still open** — bounded pagination (or another durable recovery mechanism) for the historical rebuild, and the NIP-01 event-id tie-break. On the tie-break, section 2.1's counter-argument — `created_at` is monotonic per client, so equal timestamps will not appear from a well-behaved client — does not reach the case: `d=notify` is a public replaceable event any NIP-51 client can publish, so the ordering guarantee cannot rest on our client's behaviour.
- **The fan-out decision.** `handle_content_event` awaits each recipient serially (`event_handler.rs:441-472`), `send_batch` loops tokens serially (`fcm_sender.rs:286`), and `run()` awaits `route_event` before pulling the next event. Risks still records only the `SMEMBERS`. A cap, bounded concurrency, or an explicit statement of the accepted fan-out size all work — but one of them should be written down.
- **Ingest authorization.** The new cap bounds a single list; it does not bound the number of distinct `notify_watchers:*` keys creatable from fresh keypairs, none of which can ever produce a push.
- **Nothing removes a user from the index**, including on deregistration.
- **`video_recipient_claim_key` has no notification-type component** (`event_handler.rs:533-541`), so a NewPost push burns the year-long claim a later Mention on an edit would have used.
- **`d=block` as precedent** — mobile is migrating that list to the standard kind 10000 (`content_blocklist_repository.dart:32`) — and whether the `d` value should be app-scoped, given kind 30000's `d` namespace is user-facing and shared with other clients.
- The design still reads `Status: ready for implementation`.

## Two things I could not verify

**The packet's premise is not reviewable as it stands.** Section 1 supersedes the design's task-4 blocker and Prompt 4 forbids a backfill, both resting on `design/bell-notifications` being built and tested but unpushed. That branch is not on origin, and `newPostsEnabled`, `publishedKindsSchemaVersion`, `Nip51PeopleListCodec.notifyDTag` and `reservedDTags` are absent from every reachable ref. What is reachable disagrees with it: `notificationKindFromPushType` has exactly five lowercase cases and no `newPost`, `NotificationPreferences` has five booleans, and the codec reserves only `blockedDTag`. Pushing that branch — a draft PR is enough — would make the whole contract checkable; until then I would not treat task 4 as resolved. Relatedly, the packet overriding the design in three places, rather than the design being corrected, will be hard on whoever implements from both.

**This repo is public; the platform IaC repo is private.** My first topology commit quoted an internal cluster hostname, namespace, logical database index and private-repo file paths into it. `3304efa` removes them and keeps the technical point — my mistake, flagged rather than quietly rewritten. Line 178 still cites a private-repo file path in your own replica-count justification, which I left alone: whether that stays is your call, but worth deciding deliberately rather than inheriting.

## Validation

Docs-only, so no code paths are exercised by the change. CI green on `3304efa`: `test` pass, `semantic-pull-request` pass, CLA pass, deploy jobs skipping as expected. Fast-forward push — no force, no rebase, no history rewrite.

Requesting changes rather than approving, since the two blocking items and the fan-out decision are yours to make. Happy to push follow-ups once the mechanisms are settled, and push back on any of the above if I have read it wrong.

PR #35, opened 2026-07-30
feat(push): deliver new-post pushes to bell subscribers

### LINKED ISSUE #33: feat(push): new-post ("bell") notifications for followed creators

rabble:
## Summary

Implements new-post ("bell") push notifications: a user gets a push when a creator they subscribed to publishes a kind 34236 video.

Subscriptions come from a public NIP-51 kind 30000 `d=notify` list. The service keeps a `creator -> subscribers` reverse index in Redis, rebuilds it from historical notify lists on startup, and on each incoming kind 34236 notifies the author's watchers, rate-limited to one push per `(subscriber, creator)` per hour.

No new FCM payload fields. `insert_video_reference_fields` already emits the routing target the client needs.

## Motivation

Existing notifications are "someone acted on your content", resolved from tags on the trigger event. New-post pushes invert that: the recipient subscribed to a creator's output, and the trigger event does not name the recipients. That requires a reverse index instead of tag-walking.

## Linked Issue

Closes #33. Design plan is #34 (docs-only, separate review). This PR targets `main` and does not stack on it.

## Cross-Repo Dependency

The mobile half is not merged yet. `divine-mobile` PR #6517 adds the bell UI, publishes the `d=notify` list, includes kind 34236 in push preferences, republishes preferences after the schema change, and routes the `newPost` push type.

This service can deploy before mobile as a dark backend change, but the feature should not be called user-visible until #6517 lands and ships behind its feature flag.

## Decisions Worth Review

- Notify lists are exempt from the 7-day replay horizon because replaceable subscription state can remain current for months.
- Notify lists are exempt from the per-event claim because they send no push and `replace_notify_subscriptions` is idempotent by `created_at:event_id`.
- Historical notify-list replay pages backward with `until`, using `notify_list_history_limit` as the page size.
- The Redis reverse-index update is one Lua diff-and-apply script. It preserves a recoverable write order and exact event replays reassert the reverse index.
- Video coordinate dedup keys are type-scoped, but legacy untyped keys are read for rollout compatibility.
- A delivered mention satisfies the bell record for the same video; a delivered bell does not satisfy a later mention.
- A rate-limited bell still records the video coordinate so later edits do not announce the suppressed video as new.
- If a watcher is also mentioned but muted mentions while allowing bells, the send path falls back to the enabled `NewPost` delivery.

## Validation

- `cargo fmt --check`
- `cargo check`
- `cargo test --lib` (93 passed)
- `cargo test --test notify_subscriptions_test -- --nocapture`
- `cargo test --test event_claim_test -- --nocapture`

Local note: this machine has no Redis listener on 6379, so suites with skip guards print their "Redis not available" probe lines and the non-skipping `preferences_test` Redis cases fail locally with bb8 connection timeouts. CI provisions Redis and should be the source of truth for the full `cargo test` run.

## Rollout Notes

New config is defaulted:

| Setting | Default |
|---|---|
| `new_post_rate_limit_secs` | `3600` |
| `notify_list_max_creators` | `1000` |
| `notify_list_history_limit` | `5000` |

`34236` is added to `default_preferences.kinds`, so new users get bells enabled by default once mobile can publish notify lists and updated preferences.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -75,7 +75,7 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 - `dedup:{event_id}` - Per-event processing claim with TTL. Notify lists are exempt: they are idempotent by `created_at`, so the claim prevents nothing and a failed handler would otherwise strand the list for the TTL
 - `dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify. Scoped by notification type so a bell does not suppress a later mention on the same video. The scoping is one-directional: a delivered mention also writes the `newPost` record, because naming the video already tells the recipient it exists
 - `notify_subs:{subscriber}` - Creators this user has belled
-- `notify_subs_ts:{subscriber}` - `created_at` of the last applied notify list (out-of-order guard)
+- `notify_subs_ts:{subscriber}` - `created_at:event_id` of the last applied notify list (out-of-order guard). The id resolves a `created_at` tie by NIP-01's lowest-id rule; a bare integer from an earlier build still reads as a timestamp with no known id
 - `notify_watchers:{creator}` - Subscribers watching this creator (hot read path)
 - `notify_rate:{subscriber}:{creator}` - New-post rate-limit window marker
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -65,11 +65,19 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 - **3079**: Register push token (encrypted)
 - **3080**: Deregister push token (encrypted)
 - **3083**: Update notification preferences
+- **30000**: NIP-51 people list; `d=notify` carries new-post ("bell") subscriptions (public, unencrypted)
 
 ### Redis Keys
-- `divine:token:{pubkey}` - User's FCM token
-- `divine:preferences:{pubkey}` - User's notification preferences (JSON)
-- `divine:dedup:{pubkey}:{event_id}` - Deduplication with TTL
+- `user_tokens:{pubkey}` - Set of FCM tokens registered for a pubkey
+- `token_to_pubkey` - Hash mapping a token back to its owner
+- `stale_tokens` - Sorted set of token timestamps, for cleanup
+- `user_preferences:{pubkey}` - User's notification preferences (JSON)
+- `dedup:{event_id}` - Per-event processing claim with TTL
+- `dedup:{kind}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify
+- `notify_subs:{subscriber}` - Creators this user has belled
+- `notify_subs_ts:{subscriber}` - `created_at` of the last applied notify list (out-of-order guard)
+- `notify_watchers:{creator}` - Subscribers watching this creator (hot read path)
+- `notify_rate:{subscriber}:{creator}` - New-post rate-limit window marker
 
 ### Notification Types
 | Type | Kind | Description |
@@ -79,6 +87,7 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 | Follow | 3 | New followers |
 | Mention | 1 | Notes mentioning user |
 | Repost | 16 | Reposts of user's notes |
+| NewPost | 34236 | A belled creator published a video (recipients from `notify_watchers`, not `p` tags) |
 
 ## Unblocking Workflow
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,7 +72,7 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 - `token_to_pubkey` - Hash mapping a token back to its owner
 - `stale_tokens` - Sorted set of token timestamps, for cleanup
 - `user_preferences:{pubkey}` - User's notification preferences (JSON)
-- `dedup:{event_id}` - Per-event processing claim with TTL
+- `dedup:{event_id}` - Per-event processing claim with TTL. Notify lists are exempt: they are idempotent by `created_at`, so the claim prevents nothing and a failed handler would otherwise strand the list for the TTL
 - `dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify. Scoped by notification type so a bell does not suppress a later mention on the same video
 - `notify_subs:{subscriber}` - Creators this user has belled
 - `notify_subs_ts:{subscriber}` - `created_at` of the last applied notify list (out-of-order guard)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,7 +73,7 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 - `stale_tokens` - Sorted set of token timestamps, for cleanup
 - `user_preferences:{pubkey}` - User's notification preferences (JSON)
 - `dedup:{event_id}` - Per-event processing claim with TTL
-- `dedup:{kind}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify
+- `dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify. Scoped by notification type so a bell does not suppress a later mention on the same video
 - `notify_subs:{subscriber}` - Creators this user has belled
 - `notify_subs_ts:{subscriber}` - `created_at` of the last applied notify list (out-of-order guard)
 - `notify_watchers:{creator}` - Subscribers watching this creator (hot read path)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,7 +73,7 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 - `stale_tokens` - Sorted set of token timestamps, for cleanup
 - `user_preferences:{pubkey}` - User's notification preferences (JSON)
 - `dedup:{event_id}` - Per-event processing claim with TTL. Notify lists are exempt: they are idempotent by `created_at`, so the claim prevents nothing and a failed handler would otherwise strand the list for the TTL
-- `dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify. Scoped by notification type so a bell does not suppress a later mention on the same video
+- `dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify. Scoped by notification type so a bell does not suppress a later mention on the same video. The scoping is one-directional: a delivered mention also writes the `newPost` record, because naming the video already tells the recipient it exists
 - `notify_subs:{subscriber}` - Creators this user has belled
 - `notify_subs_ts:{subscriber}` - `created_at` of the last applied notify list (out-of-order guard)
 - `notify_watchers:{creator}` - Subscribers watching this creator (hot read path)
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ The service subscribes to trigger events on its relay and notifies the tagged re
 | Reply | 1 | Reply to a user's note |
 | Mention | 1 | Note mentioning a user |
 | Repost | 16 | Repost of a user's note (NIP-18) |
+| New post | 34236 | A creator the user subscribed to ("belled") published a video |
+
+New-post notifications are the one type not anchored to a `p` tag on the trigger event. Recipients come from the subscriber's own NIP-51 list (kind 30000, `d=notify`), so the service resolves them from a Redis reverse index rather than from the video. They are rate-limited to one push per (subscriber, creator) per hour; the in-app feed is not throttled. See [the protocol doc](docs/nip-xx-push-notifications.md) for the list shape.
 
 Divine video comments are NIP-22 `kind:1111` and notify both the root video author and the direct parent author (deduplicated when they coincide). Follows (kind 3) are defined in the protocol and subscribed to, but new-follow notifications are **not currently emitted** — that requires diffing contact-list state, which is not yet implemented.
 
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The service subscribes to trigger events on its relay and notifies the tagged re
 | Repost | 16 | Repost of a user's note (NIP-18) |
 | New post | 34236 | A creator the user subscribed to ("belled") published a video |
 
-New-post notifications are the one type not anchored to a `p` tag on the trigger event. Recipients come from the subscriber's own NIP-51 list (kind 30000, `d=notify`), so the service resolves them from a Redis reverse index rather than from the video. They are rate-limited to one push per (subscriber, creator) per hour; the in-app feed is not throttled. See [the protocol doc](docs/nip-xx-push-notifications.md) for the list shape.
+New-post notifications are the one type not anchored to a `p` tag on the trigger event. Recipients come from the subscriber's own NIP-51 list (kind 30000, `d=notify`), so the service resolves them from a Redis reverse index rather than from the video. They are rate-limited to one push per (subscriber, creator) per hour, and fan-out is paged and delivered with bounded concurrency so one popular creator cannot force one unbounded Redis read or sequential delivery loop. The in-app feed is not throttled. See [the protocol doc](docs/nip-xx-push-notifications.md) for the list shape.
 
 Divine video comments are NIP-22 `kind:1111` and notify both the root video author and the direct parent author (deduplicated when they coincide). Follows (kind 3) are defined in the protocol and subscribed to, but new-follow notifications are **not currently emitted** — that requires diffing contact-list state, which is not yet implemented.
 
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -16,7 +16,7 @@ service:
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
   notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list can't stall Redis
-  notify_list_history_limit: 5000  # Safety valve on the unbounded historical notify-list query
+  notify_list_history_limit: 5000  # Page size for historical notify-list replay
 
 redis:
   url: "redis://127.0.0.1:6379"
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -14,6 +14,8 @@ service:
   process_window_days: 7
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
+  new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
+  notify_list_history_limit: 5000  # Safety valve on the unbounded historical notify-list query
 
 redis:
   url: "redis://127.0.0.1:6379"
@@ -49,6 +51,7 @@ notification:
       - 7      # Reactions/likes
       - 16     # Reposts
       - 30023  # Long-form content
+      - 34236  # Videos from subscribed creators (new-post "bells")
 
 server:
   listen_addr: "0.0.0.0:8000"
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -15,8 +15,10 @@ service:
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
-  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list can't stall Redis
+  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list cannot stall Redis
   notify_list_history_limit: 5000  # Page size for historical notify-list replay
+  new_post_fanout_page_size: 1000  # Watcher SSCAN page size for one video
+  new_post_delivery_concurrency: 50  # Concurrent NewPost deliveries per page
 
 redis:
   url: "redis://127.0.0.1:6379"
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -15,6 +15,7 @@ service:
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
+  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list can't stall Redis
   notify_list_history_limit: 5000  # Safety valve on the unbounded historical notify-list query
 
 redis:
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -15,8 +15,10 @@ service:
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
-  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list can't stall Redis
+  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list cannot stall Redis
   notify_list_history_limit: 5000  # Page size for historical notify-list replay
+  new_post_fanout_page_size: 1000  # Watcher SSCAN page size for one video
+  new_post_delivery_concurrency: 50  # Concurrent NewPost deliveries per page
   # allowed_pubkeys: [] — empty means no restriction (production default)
 
 redis:
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -15,6 +15,7 @@ service:
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
+  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list can't stall Redis
   notify_list_history_limit: 5000  # Safety valve on the unbounded historical notify-list query
   # allowed_pubkeys: [] — empty means no restriction (production default)
 
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -14,6 +14,8 @@ service:
   process_window_days: 7
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
+  new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
+  notify_list_history_limit: 5000  # Safety valve on the unbounded historical notify-list query
   # allowed_pubkeys: [] — empty means no restriction (production default)
 
 redis:
@@ -50,6 +52,7 @@ notification:
       - 7      # Reactions/likes
       - 16     # Reposts
       - 30023  # Long-form content
+      - 34236  # Videos from subscribed creators (new-post "bells")
 
 server:
   listen_addr: "0.0.0.0:8000"
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -16,7 +16,7 @@ service:
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
   notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list can't stall Redis
-  notify_list_history_limit: 5000  # Safety valve on the unbounded historical notify-list query
+  notify_list_history_limit: 5000  # Page size for historical notify-list replay
   # allowed_pubkeys: [] — empty means no restriction (production default)
 
 redis:
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -190,7 +190,85 @@ Users can optionally send a Kind 3083 event to control which notification types 
 { "kinds": [1, 3, 7, 16] }
 ```
 
-This is a list of event kinds the user wants notifications for. If no preferences are set, the service uses defaults: text notes (1), follows (3), reactions (7), reposts (16), and long-form content (30023).
+This is a list of event kinds the user wants notifications for. If no preferences are set, the service uses defaults: text notes (1), follows (3), reactions (7), reposts (16), long-form content (30023), and videos from subscribed creators (34236).
+
+## New-post subscriptions ("bells")
+
+Every other notification type is triggered by someone acting on the recipient's
+content, so recipients are read off the trigger event's `p` tags. New-post
+notifications invert that: the recipient subscribed to a creator's output, and
+the trigger event says nothing about who wants it.
+
+### Source of truth
+
+The subscription list is a public NIP-51 people list published by the client,
+identified by a reserved `d` tag:
+
+```json
+{
+  "kind": 30000,
+  "tags": [
+    ["d", "notify"],
+    ["title", "Notify"],
+    ["p", "<creator-pubkey-hex>"]
+  ]
+}
+```
+
+It is replaceable and unencrypted — the service has to be able to read it, so
+there is no decryption step, unlike the kind 3079/3080/3083 control events. Any
+kind 30000 arriving without exactly `d=notify` is ignored.
+
+### Ingestion
+
+`handle_notify_list_update` is routed **before** the control-event block and
+deliberately outside its `p`-tag gate: these events are addressed to the world,
+not to this service.
+
+Two properties are load-bearing:
+
+- **Notify lists are exempt from the replay horizon.** A list published three
+  months ago and never touched since is still the user's current subscription
+  set, so `is_notify_list` carves it out of the `is_event_too_old` check. The
+  historical query is likewise unbounded in time, with
+  `notify_list_history_limit` as the size safety valve instead. Without
+  historical replay, a restart against a fresh Redis silently drops every bell
+  until each user republishes.
+- **An empty `p` list is legitimate**, not malformed. It means the user unbelled
+  everyone, and it must clear the forward set and remove them from every reverse
+  index.
+
+`replace_notify_subscriptions` applies the diff in a single Lua script keyed on
+the subscriber, because `notify_subs` and `notify_watchers` are two views of one
+relation and must move together. The script re-checks the stored `created_at`
+internally — a relay can deliver an older replacement after a newer one, and an
+advisory check in the caller would still race.
+
+The script writes `notify_watchers:*` keys that are not declared in `KEYS`, so it
+is **not Redis Cluster safe**. This deployment uses single-instance Redis; moving
+to Cluster requires resharding into one call per creator slot or a hash-tagged
+key layout.
+
+### Delivery
+
+On each incoming kind 34236, `video_notification_targets` unions the video's
+mention targets with `SMEMBERS notify_watchers:{author}`.
+
+**Mention wins on overlap.** A user who both watches the creator and is mentioned
+in the video gets one push, typed `mention`, because that is the more specific
+signal.
+
+Delivery is capped at one push per (subscriber, creator) per
+`new_post_rate_limit_secs`. The window is opened only on a *delivered* push
+(check-then-set-on-success, mirroring the video-coordinate dedup) so a failed FCM
+send does not burn the user's hour. The cost is that two replicas handling
+different videos from the same creator in the same instant can both pass the
+check and double-send — rare, bounded, and preferable to silently eating an hour
+of notifications on an FCM blip.
+
+The rate limit is push-only. The in-app feed shows every post from belled
+creators, so a user who receives one push for a six-post burst opens the app and
+sees all six. That is intended.
 
 ## Redis Keys
 
@@ -201,4 +279,8 @@ This is a list of event kinds the user wants notifications for. If no preference
 | `stale_tokens` | Sorted Set | Token timestamps for cleanup |
 | `dedup:{event_id}` | String | Deduplication lock with TTL |
 | `dedup:34236:{owner}:{d-tag}:{recipient}` | String | Successful per-recipient video delivery, retained for the configured coordinate TTL (one year by default) |
-| `divine:preferences:{pubkey}` | String | JSON notification preferences |
+| `user_preferences:{pubkey}` | String | JSON notification preferences |
+| `notify_subs:{subscriber}` | Set | Creators this user has belled. Diffed against each incoming replacement list. |
+| `notify_subs_ts:{subscriber}` | String | `created_at` of the last applied notify list. Guards against out-of-order relay delivery of a replaceable event. |
+| `notify_watchers:{creator}` | Set | Subscribers watching this creator. The hot read path — one `SMEMBERS` per incoming video. |
+| `notify_rate:{subscriber}:{creator}` | String | New-post rate-limit window marker, TTL `new_post_rate_limit_secs` (one hour by default). |
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -258,6 +258,26 @@ relation and must move together. The script re-checks the stored `created_at`
 internally — a relay can deliver an older replacement after a newer one, and an
 advisory check in the caller would still race.
 
+The write order inside the script is deliberate. Redis runs a script without
+interleaving anything else, but it does not roll one back, so a script that dies
+partway keeps what it already wrote. Removals therefore clear the reverse index
+before the forward one and additions write the forward index first, which keeps
+`notify_subs` a *superset* of the true relation at every intermediate step. That
+matters because `notify_subs` is the only record of which `notify_watchers:*`
+keys name a subscriber, and removals are computed from it: a superset is
+reconciled by the next list, while a forward index that is missing entries the
+reverse index still holds cannot be repaired by anything the user can publish.
+Do not "simplify" the diff back into a `DEL` and rebuild.
+
+This covers the script failing on its own. It does not cover the index being
+lost some other way, and the startup replay only partly does. A *total* loss
+rebuilds: `notify_subs_ts` goes with everything else, so every replayed list
+passes the guard and re-applies. A *partial* loss does not: if the index is gone
+but `notify_subs_ts` survives, the replay re-fetches the same replaceable event
+and the guard rejects it on exact-id match, leaving those bells dead until the
+user next publishes a genuinely newer list. Deleting `notify_subs_ts` for the
+affected subscribers is currently the only lever.
+
 Ties on `created_at` resolve by NIP-01's rule, retaining the lowest event id.
 The protocol requires clients to publish the complete list on every change, so
 belling two creators in quick succession produces two full-list publishes that
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -270,13 +270,19 @@ reverse index still holds cannot be repaired by anything the user can publish.
 Do not "simplify" the diff back into a `DEL` and rebuild.
 
 This covers the script failing on its own. It does not cover the index being
-lost some other way, and the startup replay only partly does. A *total* loss
-rebuilds: `notify_subs_ts` goes with everything else, so every replayed list
-passes the guard and re-applies. A *partial* loss does not: if the index is gone
-but `notify_subs_ts` survives, the replay re-fetches the same replaceable event
-and the guard rejects it on exact-id match, leaving those bells dead until the
-user next publishes a genuinely newer list. Deleting `notify_subs_ts` for the
-affected subscribers is currently the only lever.
+lost some other way, and the startup replay covers that in one direction only.
+A *total* loss rebuilds: `notify_subs_ts` goes with everything else, so every
+replayed list passes the guard and re-applies. A lost *reverse* index rebuilds
+as well, even with `notify_subs_ts` intact, because an exact-id replay is
+applied rather than rejected (see the tie-break note below) and re-asserts every
+`notify_watchers:*` entry the list implies.
+
+The mirror case is the one the replay cannot repair: if `notify_subs` is lost
+while the reverse index survives, the removal loop has nothing to diff against,
+so watcher entries for creators the subscriber has since unbelled stay behind
+and keep delivering. Republishing does not clear them — a new list only adds —
+which is the same asymmetry the write ordering above exists to avoid creating.
+Removing those `notify_watchers:*` entries directly is the only lever.
 
 Ties on `created_at` resolve by NIP-01's rule, retaining the lowest event id.
 The protocol requires clients to publish the complete list on every change, so
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -258,6 +258,15 @@ relation and must move together. The script re-checks the stored `created_at`
 internally — a relay can deliver an older replacement after a newer one, and an
 advisory check in the caller would still race.
 
+Ties on `created_at` resolve by NIP-01's rule, retaining the lowest event id.
+The protocol requires clients to publish the complete list on every change, so
+belling two creators in quick succession produces two full-list publishes that
+can share a second; resolving those by arrival order would leave this service
+holding a different list than the relay does, permanently. Watch the direction
+when reading the script: an equal-timestamp event is applied only when its id
+sorts *below* the stored one, which reads backwards from "newer wins", and an
+exact replay is still rejected.
+
 Because the script runs as one blocking unit and Redis is single-threaded, the
 creator list is bounded by `notify_list_max_creators` before it reaches Redis —
 otherwise one user with an absurd number of bells stalls the instance for
@@ -317,6 +326,6 @@ sees all six. That is intended.
 | `dedup:34236:{type}:{owner}:{d-tag}:{recipient}` | String | Successful per-recipient video delivery, retained for the configured coordinate TTL (one year by default). `{type}` is the notification type (`newPost`, `mention`), so a bell and a mention for the same video coordinate keep independent records. A delivered mention writes both records, since naming the video already tells the recipient it exists; a delivered bell writes only its own |
 | `user_preferences:{pubkey}` | String | JSON notification preferences |
 | `notify_subs:{subscriber}` | Set | Creators this user has belled. Diffed against each incoming replacement list. |
-| `notify_subs_ts:{subscriber}` | String | `created_at` of the last applied notify list. Guards against out-of-order relay delivery of a replaceable event. |
+| `notify_subs_ts:{subscriber}` | String | `created_at:event_id` of the last applied notify list. Guards against out-of-order relay delivery of a replaceable event, and carries the id so a `created_at` tie resolves by NIP-01's lowest-id rule. A bare integer written by an earlier build is read as a timestamp with no known id, which only makes the guard more conservative. |
 | `notify_watchers:{creator}` | Set | Subscribers watching this creator. The hot read path — one `SMEMBERS` per incoming video. |
 | `notify_rate:{subscriber}:{creator}` | String | New-post rate-limit window marker, TTL `new_post_rate_limit_secs` (one hour by default). |
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -291,7 +291,7 @@ sees all six. That is intended.
 | `token_to_pubkey` | Hash | Reverse mapping from token to owner pubkey |
 | `stale_tokens` | Sorted Set | Token timestamps for cleanup |
 | `dedup:{event_id}` | String | Deduplication lock with TTL |
-| `dedup:34236:{owner}:{d-tag}:{recipient}` | String | Successful per-recipient video delivery, retained for the configured coordinate TTL (one year by default) |
+| `dedup:34236:{type}:{owner}:{d-tag}:{recipient}` | String | Successful per-recipient video delivery, retained for the configured coordinate TTL (one year by default). `{type}` is the notification type (`newPost`, `mention`), so a bell and a mention for the same video coordinate keep independent records |
 | `user_preferences:{pubkey}` | String | JSON notification preferences |
 | `notify_subs:{subscriber}` | Set | Creators this user has belled. Diffed against each incoming replacement list. |
 | `notify_subs_ts:{subscriber}` | String | `created_at` of the last applied notify list. Guards against out-of-order relay delivery of a replaceable event. |
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -37,6 +37,7 @@ sequenceDiagram
 | 3079 | Client → Relay → Service | Register FCM push token (NIP-44 encrypted) |
 | 3080 | Client → Relay → Service | Deregister push token (NIP-44 encrypted) |
 | 3083 | Client → Relay → Service | Update notification preferences (optional) |
+| 30000 | Client → Relay → Service | NIP-51 people list; `d=notify` carries new-post ("bell") subscriptions. Public and unencrypted, and addressed to the world rather than `p`-tagged to the service |
 
 See [NIP-XX Push Notifications](nip-xx-push-notifications.md) for the full protocol specification.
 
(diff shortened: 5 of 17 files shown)


mbradley (changes requested):
# PR #35 — drafted review body (verdict: CHANGES_REQUESTED)

Reviewer takeover. I pushed five commits covering the findings whose remedy was
mechanical, and left the four that need your call as questions below. Everything
I touched is in `src/` and `tests/` — no config, no CI, no workflows.

Gates checked first: org member with write access, in-repo branch rather than a
fork, non-draft, mapping readable (`divine-push-service` is unmapped, so the
default rule applies), changed paths inside the low-risk boundary. One commit per
finding so you can revert any of them independently.

Validation: `cargo test` 109 passed / 0 failed with Redis on 6379 (confirmed the
8 `notify_subscriptions_test` cases execute rather than skip), `cargo clippy
--all-targets --all-features -- -D warnings` clean, `cargo fmt --check` clean.

## Pushed

**1. A watcher-lookup failure was taking the mentions down with it.**
`video_notification_targets` propagated the `get_notify_watchers` error, so a
transient Redis blip discarded the video's *mention* pushes too — targets that
came from the event's own tags and needed no lookup at all. Bells are enrichment
on top of mentions; they should cost only themselves when they fail. Degrades to
no bells and logs. The fallback is a pure function over the lookup `Result` so
it is testable without an `AppState`, same rationale as `merge_watcher_targets`.

**2. The replay-horizon exemption had no test.** It is the correction this PR
leads with, and its failure mode is silent — bells set more than seven days ago
just never rebuild and nothing logs an error. The plan asked for a boundary test
either side of the horizon (#34, task 1) and it did not land. Extracted the guard
out of `run()` into `is_beyond_replay_horizon` so a test can reach it, and
covered both sides. Verified the test fails when the exemption is removed.

**3. `is_notify_list` checked only the kind, not the `d` tag.** So every kind
30000 on the relay bypassed the horizon. `handle_notify_list_update` explicitly
declines to trust the relay's `#d` narrowing — "a buggy or hostile relay can send
any kind 30000 it likes" — and the horizon bypass sat outside that defence, so
arbitrary old people lists got through to claim a dedup key before being dropped.
Same check, applied in both places.

**4. The live notify-list subscription replayed history on every reconnect.** It
reused `notify_list_filter(notify_list_history_limit)` verbatim. Per NIP-01 a
filter's `limit` applies to the initial stored-event query, so each subscribe —
including nostr-sdk's automatic resubscribe after a relay drop — asked for up to
5000 stored lists the historical pass had already applied, each costing a channel
send and a Redis claim round-trip to reach a no-op. Split into a history filter
(unbounded in time, `limit` as the safety valve) and a live filter with the same
one-hour `since` the main live filter uses. This bounds the replay rather than
eliminating it, which matches how the existing subscription behaves.

**5. The event log lost its notification type.** `notification_type` was dropped
from the "Processing notification event" line because targets became mixed-type,
and nothing replaced it. That was the one line saying what an event produced, and
NewPost — fan-out plus a rate limit — is exactly what you would be grepping for.
Logs the per-type breakdown alongside the count.

## Questions — yours to decide, not mine to patch

**A. Same-second replacement resolves by arrival order.** The Lua guard is
`stored >= incoming → reject`, and `test_replaying_the_same_created_at_is_ignored`
now asserts that as correct. NIP-01 says otherwise: same-timestamp replaceable
events resolve by lowest event id (`nips/01.md:101`). A bell and an unbell in the
same second therefore resolve by whichever the relay delivers first, and the test
locks that in. @dcadenas requested this on #34 and it is unchanged here; flagging
it because the new test makes the current behaviour look deliberate. Storing the
winning event id next to `created_at` and tie-breaking inside the script is a
schema change, so it is yours rather than something I would push.

**B. `try_claim_event` turns a transient failure into up to 7 days of lost
subscriptions.** `run()` claims the event before routing. For every other handler
that is fine — losing one push is invisible. The notify-list handler builds
*persistent* state, so if `replace_notify_subscriptions` errors, the error is
logged, the claim stands, and the historical replay on the next restart skips the
event as already-claimed. That user's bells stay dark until the dedup key expires
and something restarts. #34 flagged this interaction as unresolved ("the plan
should say how the rebuild is meant to interact with that") and it is still open.
Options: (a) exempt kind 30000 from `try_claim_event` — the Lua script is already
atomic and idempotent by `created_at`, so the claim buys nothing here;
(b) release the claim when the handler errors; (c) accept it and say so in a
comment. I lean (a), symmetric with the horizon exemption you already made.

**C. The sender display name is resolved once per recipient.** `create_fcm_payload`
computes `sender_name` from `event.pubkey` — identical for every target of the
event — inside the per-recipient loop. Each call is a Redis GET on
`profile_cache:{author}`; on a miss it is a relay `fetch_events`, and misses are
not negatively cached, so a creator with no kind-0 metadata produces one relay
round-trip *per watcher*. With mentions that was 1-3 recipients. With bells it is
every watcher of a creator. `format_content_for_push` is event-scoped in the same
way. Hoisting both out of the loop is the obvious fix, but where it lands depends
on what you do about the serial fan-out @realmeylisdev raised on #34, so I have
not touched it. Fix it here, or fold it into the fan-out work?

**D. Nothing ever removes a subscriber.** `notify_subs:*`, `notify_subs_ts:*` and
`notify_watchers:*` have no TTL, `handle_deregistration` does not touch them, and
`cleanup_service.rs` is unchanged. A user who deregisters leaves a permanent
record of which creators they watched. @realmeylisdev raised this on #34 as a
retention question; it is unchanged in the code. Deregistration cleanup, a GC
sweep, or documented-and-accepted?

## Still open from #34, unchanged here

Not re-arguing these, just noting they carried into the implementation and have
no disposition yet:

- A NewPost push burns `34236:{author}:{d}:{recipient}` for a year, so a later
  edit that adds a `p` tag mentioning that watcher is suppressed — belling a
  creator quietly costs you mention notifications from them (@realmeylisdev).
- Ingest has no authorization and no bound on distinct subscribers; the creator
  cap bounds one list, not the number of pubkeys that can create one
  (@realmeylisdev).
- `d="notify"` is a shared user-facing namespace, so an unrelated follow set
  named "notify" from another client reads as a bell list (@realmeylisdev).

Happy to take any of those on once you have decided which way they go.

realmeylisdev (changes requested):
Reviewer takeover. Three commits pushed, one per finding so any can be reverted
alone. Gates checked first: org member with write access, in-repo branch rather
than a fork, non-draft, `divine-push-service` unmapped so the default rule
applies, and every changed path inside `src/` and docs — no workflows, no CI, no
deployment config.

## First: B and C are not on this branch

The comment above says B and C are "implemented and verified, one commit each."
Neither is on `feat/new-post-notifications`.

When I picked this up the head was `5556912` with nine commits — four yours, all
predating @mbradley's review, and five of his. `run()` still claimed every event
unconditionally, and `create_fcm_payload` still resolved `sender_name` per
recipient. The test count agrees: the branch defined exactly 109 tests, matching
the 109 in his review rather than the 110 in yours. PR #36 branches from
`4a5a7a8` and is a different fix.

So that work exists somewhere local but never left the machine. Worth checking
before @dcadenas rules on anything, because the branch being described and the
branch on GitHub are not the same branch.

I have pushed C, and the claim-key fix. I left B — the claim exemption — alone,
since your unpushed commit presumably has it and I did not want to collide with
it.

## Pushed

**1. `perf(push): resolve the sender name once per event, not per recipient` (`a62f67b`)**

This is C, and it is worse than the review describes. On a profile-cache miss
`get_display_name` does a relay `fetch_events` with a five-second timeout, and
`fetch_profiles_batch` only writes back profiles it actually found — a miss is
never negatively cached. So a creator with no kind-0 event pays that round-trip
on *every* call, and `create_fcm_payload` runs per recipient. With `send_batch`
a sequential loop running on the event-handler loop, N watchers of a
profile-less creator can hold that loop for up to 5N seconds with every other
user's notifications queued behind it.

That is a head-of-line stall any creator without a profile can trigger, not
throughput tuning, which is why I did not fold it into #37. Both values are now
resolved once in `handle_content_event` and passed down. The content parse is
skipped unless some target actually renders the body, so a pure bell fan-out no
longer pays for mention parsing it never uses. `create_fcm_payload` is left free
of I/O, which is what lets the new tests reach it directly.

Behaviour is unchanged, including the fallback to raw content when the parse is
skipped or fails.

**2. `fix(push): scope the video dedup claim to the notification type` (`fb5f667`)**

Your lean, implemented. The key becomes
`dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}`.

I reproduced the failure before fixing it rather than reasoning about it:
reverting just the key format makes the new
`test_a_bell_does_not_suppress_a_later_mention_on_the_same_video` fail with one
push delivered where two are owed — the bell swallowing the mention, exactly as
you described. Both Redis key tables are updated.

Worth noting for deploy: existing `dedup:34236:*` keys are orphaned by the new
format and age out on their own TTL. During the transition a recipient already
notified under the old key can receive one further push for that coordinate.
One-time and bounded.

**3. `fix(push): drop self-targets before resolving the event-scoped copy` (`fe56275`)**

A regression commit 1 introduced, caught on re-reading the diff. Several of the
`find_*` helpers return the author when an event `p`-tags its own sender, and
the send loop skipped those one recipient at a time. That was free while the
payload was built per recipient; it is not once the copy is resolved before the
loop, because a self-only event then paid for a profile lookup it never used.
Filtering at construction also makes `recipient_count` and the per-type
breakdown describe real deliveries rather than candidates.

## Validation

`cargo test` 117 passed / 0 failed against Redis on 6379 — up from 109, the
eight extra being the new coverage below. `cargo clippy --all-targets
--all-features -- -D warnings` clean, `cargo fmt --check` clean. CI green on
`fe56275`.

| Behaviour | Test |
|---|---|
| Bell and mention keep independent records | `test_video_recipient_claim_key_separates_notification_types` |
| A delivered bell does not suppress a later mention | `test_a_bell_does_not_suppress_a_later_mention_on_the_same_video` |
| Only body-rendering types need the content | `test_only_body_rendering_types_need_the_event_content` |
| Payload uses the event-scoped sender name | `test_payload_uses_the_event_scoped_sender_name` |
| Mention body prefers the resolved content | `test_mention_body_prefers_the_resolved_content` |
| Mention body falls back to raw content | `test_mention_body_falls_back_to_raw_content` |
| Self-targets are dropped | `test_deliverable_targets_drops_the_author` |
| A self-only event delivers nothing | `test_deliverable_targets_empties_a_self_only_event` |

## Still yours

**A.** Agree it should be fixed, and NIP-01 is unambiguous: "In case of
replaceable events with the same timestamp, the event with the lowest id (first
in lexical order) should be retained." The relay-divergence argument stands on
its own.

The second argument in your comment does not survive the fix, though.
Lowest-id-wins does not stop a user losing a bell when two full-list publishes
land in the same second — it only decides *which* one is lost, and makes us
agree with the relay about it. Whenever the second publish sorts higher it is
still rejected, so "I belled them and it didn't stick" persists, now
nondeterministically. If that symptom is the thing worth fixing, the fix is
client-side: monotonic `created_at` across rapid republishes. Worth separating,
because the two motivations imply different follow-up work.

One implementation trap worth stating before anyone writes it: lowest-id-wins
means accepting the incoming event only when its id sorts *below* the stored
one, which reads backwards from "newer wins". And the guard still has to reject
an exact replay — same `created_at`, same id.

**D.** Agree with your lean, and the trap you name holds up: the historical
notify-list fetch runs once at startup, so deleting on deregistration would not
be repaired until a restart. Document and accept now, then do it alongside
rebuild-on-registration in #37.

Unchanged and still without a disposition: ingest authorization and the bound on
distinct subscribers, and the shared `d="notify"` namespace.

Happy to take A on once you and @dcadenas have settled the direction. It is
contained, but it is a stored-format change with a legacy read path, which is
why I left it rather than pushing it.

mbradley (commented):
# PR #35 — drafted review body (verdict: COMMENT)

Reviewer takeover, one commit. Gates re-checked against the current head: org
member with write access, in-repo branch rather than a fork, non-draft,
`divine-push-service` unmapped so the default rule applies, changed paths inside
`src/`, `tests/` and docs. No workflows, no CI, no deployment config.

## B fell through the gap between two reviews

B is the one finding on this PR that nobody was holding. Your comment reported
it implemented and verified. @realmeylisdev checked the branch, found it absent,
and deliberately left it alone rather than collide with your unpushed commit. So
both of us deferred to a commit that is not on GitHub, and the finding quietly
had no owner.

I confirmed it against `fe56275` before writing anything: `run()` still called
`try_claim_event` unconditionally at `event_handler.rs:129`, with the replay
horizon above it as the only notify-list carve-out.

Pushing it rather than asking again, because it is no longer a judgment call.
You picked option (a) explicitly, @realmeylisdev did not dissent, and the remedy
is mechanical. Revert it freely if your local version differs.

**`fix(push): exempt notify lists from the event claim`**

`requires_event_claim` sits next to `is_beyond_replay_horizon` and the claim
block in `run()` is now conditional on it. Scoped by `is_notify_list` rather than
kind alone, matching what your `769aa43` did to the horizon exemption: the
idempotency argument rests on the Lua script, and the Lua script only runs for
`d=notify`. An unrelated kind 30000 people list keeps the claim.

Worth stating because it is the obvious objection: dropping the claim does not
make the historical replay more expensive. `replace_notify_subscriptions` bails
on its first two lines when `stored >= incoming`, so a replayed list costs one
Lua eval instead of one `SET NX`. The `SMEMBERS`/`DEL`/`SADD` work is never
reached.

Docs: the ingestion section of the developer guide now lists the exemption
alongside the horizon one, and both `dedup:{event_id}` rows say why the claim is
not taken.

## Disposition of my earlier review

That review was written against `5556912` and parts of it are now stale. Closing
it out so nothing is left ambiguous:

| Finding | Where it landed |
|---|---|
| 1–5 (pushed) | On the branch, unchanged |
| B — claim strands subscriptions | Pushed here as `fcb1ef5` |
| C — sender name per recipient | Resolved by @realmeylisdev in `a62f67b`, and the head-of-line stall framing is a better account of it than mine |
| Claim-key collision (from #34) | Resolved by @realmeylisdev in `fb5f667`, reproduced before fixing |
| A — same-second resolution | Still open, yours and @dcadenas's |
| D — nothing removes a subscriber | Document and accept now, #37 later. Agreed, no objection |

So my `CHANGES_REQUESTED` is down to A, plus the two items below.

## Still without a disposition

Both carried from #34 through two reviews without an answer either way. Not
re-arguing them, just noting they are the last things on this PR with no exit:

- Ingest has no authorization and no bound on distinct subscribers. The creator
  cap bounds one list, not how many pubkeys can publish one.
- `d="notify"` is a shared user-facing namespace, so an unrelated list named
  "notify" from another client reads as a bell list.

Accepting either with a comment is a fine answer. Leaving them unanswered
through a third review is the thing worth avoiding.

## Validation

`cargo test` 123 passed / 0 failed against Redis on 6379, up from 117, the six
extra being the new coverage below. No test reported the skip-when-unavailable
path. `cargo clippy --all-targets --all-features -- -D warnings` clean,
`cargo fmt --check` clean.

I verified the new coverage can fail: making `requires_event_claim` return `true`
unconditionally turns
`test_notify_lists_are_exempt_from_the_event_claim` and
`test_the_exemption_composes_with_the_replay_horizon` red.

| Behaviour | Test |
|---|---|
| Notify lists take no claim | `test_notify_lists_are_exempt_from_the_event_claim` |
| Other kind 30000 lists still do | `test_the_exemption_does_not_cover_other_kind_30000_lists` |
| Everything that sends a push still does | `test_push_sending_events_still_claim` |
| A standing claim really would refuse the replay | `test_a_standing_claim_would_have_skipped_the_replayed_list` |
| A replayed list re-applies safely without one | `test_a_replayed_list_still_applies_without_a_claim` |
| Both loop gates let an old list through | `test_the_exemption_composes_with_the_replay_horizon` |

CI is green on `fcb1ef5`: `test` and `semantic-pull-request` both pass.

## Verdict

Leaving this as a comment rather than a verdict. @realmeylisdev's
`CHANGES_REQUESTED` already covers the current head, and A is genuinely yours
and @dcadenas's to settle, so a second one would say nothing new. Happy to
approve once A has a direction.

realmeylisdev (changes requested):
Reviewer takeover, four commits, one per finding so each reverts alone. Gates
re-checked against `fcb1ef5`: org member with write access, in-repo branch rather
than a fork, non-draft, `divine-push-service` unmapped so the default rule
applies, changed paths inside `src/`, `tests/` and docs. No workflows, no CI, no
deployment config.

## First: A was already decided, and half of that decision is still missing

Three rounds on this PR have treated A as blocked on @dcadenas. It is not.
@dcadenas reviewed #34 on 2026-07-29 with `CHANGES_REQUESTED` and has not
revisited it. That review is two requirements, verbatim:

> Add bounded pagination or an equivalent recovery strategy **and** the NIP-01
> event-id tie-break, with tests covering both cases.

So A has had a direction since before this branch existed. I have implemented it
(`60f195a`), using the storage shape @rabble specified.

The other half has never been mentioned on #35 by anyone, including me. The
historical rebuild is still a flat `limit: 5000` with no `since` and a warning
when the result count reaches the limit. On a fresh Redis with more than
`notify_list_history_limit` notify lists on the relay, the oldest lists are
dropped and those users' bells never come back until they happen to republish.

Worth knowing before choosing a strategy: `wss://relay.divine.video` advertises
`"max_limit": 5000` in its NIP-11 document, which is exactly the configured
default. Raising `notify_list_history_limit` past it does not buy more coverage —
the relay clamps — and once clamped, `count >= limit` stops being true, so the
"some subscriptions may be missing" warning goes quiet precisely when it starts
mattering. The escape hatch is not there.

I have not implemented this one. "Bounded pagination or an equivalent recovery
strategy" is at least two different designs — an `until`-cursor walk backwards,
or dropping the startup bulk fetch in favour of rebuild-on-registration, which is
where you already pointed D. That is your call and @dcadenas's, not a mechanical
remedy.

## Pushed

**1. `perf(push): resolve the event copy only when a push is actually sent` (`b5edabd`)**

My own regression from `a62f67b`, and the most expensive thing on the branch.

Hoisting `resolve_event_scoped_copy` out of the per-recipient loop was right.
Where it landed was not: above every gate in `send_notification_to_user`, so an
event pays for its copy before anything checks whether one recipient can receive
it. The token gate alone drops nearly all of them — the live subscription carries
no `#p` narrowing, so the service sees the whole firehose of kinds 1, 7, 16,
1111, 30023 and 34236, and almost no tagged user has registered for push.

On `main` those events cost nothing, because the lookup sat inside
`create_fcm_payload`, downstream of `tokens.is_empty()`. On the branch each one
costs a `profile_cache:` GET and, on a miss, a relay `fetch_events`. Misses are
never written back — `fetch_profiles_batch` caches only profiles it found — so an
author with no kind 0 pays it on every event indefinitely, on the single
sequential handler task.

I measured it rather than reasoning about it. Same branch, same input (a kind 1
whose only `p`-tagged recipient has no token), one unreachable relay in the
profile pool: **5.009 s eager, 937 µs lazy**. `Client::fetch_events` waits for
EOSE from every read relay, so one flaky profile relay among healthy ones is
enough to hit the full `query_timeout_secs`. That turns a third-party relay
wobble into a service-wide push stall, which is a strictly worse shape than the
per-recipient cost the commit set out to remove.

`LazyEventCopy` keeps the once-per-event property and defers the work to the
first recipient that clears every gate. `needs_content` is still decided over the
whole target list, so resolving on a bell rather than a mention cannot silently
downgrade the mention's body to raw content.

**2. `fix(push): stop a video mention from re-notifying as a bell after an edit` (`a10a02b`)**

Also mine, from `fb5f667`. Putting the type in the claim key made the two
directions independent. Only one of them should be.

A mention push names the video, so it already tells the recipient that video
exists — which is the entire content of a bell. A watcher who was `p`-tagged in
the original and dropped from an edit resolves to a bare `NewPost` target on the
edit, finds no `newPost` record, and is told "posted a new vine" about a video
they were pushed about already.

I reproduced it before fixing it: the new
`test_a_mention_suppresses_a_later_bell_on_the_same_video` fails on `fcb1ef5`
with two pushes delivered where one is owed. `satisfied_video_claims` makes the
carry one-directional — a delivered mention writes both records, a delivered bell
writes only its own. That is `merge_watcher_targets`'s "mention wins on overlap"
extended across edits, which the claim-key doc comment already claimed to do.

**3. `fix(push): resolve a notify-list created_at tie by NIP-01's lowest id` (`60f195a`)**

A, per the requirement above. `notify_subs_ts` now stores
`created_at:event_id`, and an equal-timestamp event applies only when its id
sorts strictly below the stored one.

Both traps are covered. The comparison reads inverted from "newer wins", and an
exact replay — same timestamp, same id — must still be rejected. No backfill: a
bare integer from an earlier build reads as a timestamp with no known id, so ties
against it stay rejected exactly as they are today, and the first list that
applies rewrites the key in the new format. There is a test for that path
specifically, since every deployed instance will be reading it.

Your second argument for A still does not survive the fix, and I would not want
it written up as though it does — lowest-id-wins does not stop a user losing a
bell when two full-list publishes land in the same second, it only makes us agree
with the relay about which one is lost. The state-divergence argument carries it
on its own.

**4. `docs: describe new-post notifications where the guide already lists types` (`a420e32`)**

The new-post sections landed as a self-contained block, so four pre-existing
tables and notes still describe the pre-PR service: the Event Kinds table has no
kind 30000 row, the Notification Types table has no NewPost row, the note under
it still says likes/comments/mentions/reposts "are the types actually delivered
today", and the payload contract enumerates the `type` values a client routes on
while omitting `newPost` and asserting the set is lowercase. That last row is
normative — divine-mobile matches the string exactly — so it was the one worth
catching. The Deduplication section also still said every event is claimed
exactly once, which `fcb1ef5` made untrue.

## Not pushed — two behaviours, both reproduced, both your call

**E. Being mentioned costs a watcher the bell they subscribed to.**

`merge_watcher_targets` types the overlap `Mention`, and the preference gate then
tests `Mention.kind()`, which is 1. A watcher with bells on and mentions off gets
**nothing** when a creator they belled posts a video that also `p`-tags them.
Confirmed against a live Redis with `kinds: [34236]`: targets resolve to
`["mention"]`, pushes delivered `0`.

"Mention wins because it is the more specific signal" is right when the mention
is deliverable. When it is muted, the specific signal has been declined and the
general one the user explicitly asked for is discarded with it. The fix is a
fallback to `NewPost` when the mention is disabled and the recipient is also a
watcher, but that means resolving preferences before or during the merge, which
is a shape change to a path #37 is about to touch. Your call on whether it is
worth it and where it lands.

**F. An edit re-announces a video the rate limit suppressed.**

The coordinate record is written only on a delivered push, and the rate-limit
check returns before it. So: creator posts V1, watcher gets a bell, window opens.
Creator posts V2 inside the window — suppressed, no record written. The hour
passes, the creator edits V2, and the watcher gets "posted a new vine" for an
hour-old video, fired by an edit. Reproduced end to end: 0 pushes for the
rate-limited video, 1 for its later edit.

Both directions are defensible — writing the record on suppression means that
video is dropped for that user permanently, which is arguably what the rate limit
already decided. Not something I would pick for you.

## Two degenerate config values, if you want them closed cheaply

Neither is reachable without a typo in `settings.yaml`, so I have left both:

- `notify_list_max_creators: 0` makes `collect_notify_creators` break on its
  first iteration, so every incoming list is read as empty and clears that
  subscriber's bells. A cap of zero silently means "delete everything".
- `new_post_rate_limit_secs: 0` makes the post-send `SETEX` error, so a
  successfully delivered bell returns `Err` and skips the coordinate write below
  it.

## Still without a disposition

Carried from #34 through three reviews, unchanged and not re-argued here:

- Ingest has no authorization and no bound on distinct subscribers. The creator
  cap bounds one list, not how many pubkeys can publish one.
- `d="notify"` is a shared user-facing namespace, so an unrelated list named
  "notify" from another client reads as a bell list.

## Validation

`cargo test` **131 passed / 0 failed** against Redis on 6379, up from 123 on
`fcb1ef5`. Re-run against a flushed Redis with the same result, and no test
reported the skip-when-unavailable path. `cargo clippy --all-targets
--all-features -- -D warnings` clean, `cargo fmt --check` clean. CI green on
`60f195a`.

Every new test was checked to fail when its subject is broken:

| Behaviour | Test | Goes red when |
|---|---|---|
| A mention suppresses a later bell | `test_a_mention_suppresses_a_later_bell_on_the_same_video` | `satisfied_video_claims` returns the identity |
| The carry is one-directional | `test_a_mention_satisfies_the_bell_record_but_not_the_reverse` | " |
| No deliverable recipient, no copy resolved | `test_a_recipient_without_tokens_never_resolves_the_event_copy` | the copy is resolved eagerly again |
| A delivered push still resolves it | `test_a_delivered_push_resolves_the_event_copy` | " |
| `needs_content` spans the whole target list | `test_needs_content_is_decided_over_the_whole_target_list` | it is decided per recipient |
| A tie retains the lower id | `test_a_created_at_tie_is_resolved_by_the_lower_event_id` | the id comparison is removed |
| A tie discards the higher id | `test_a_created_at_tie_rejects_a_higher_event_id` | " |
| A legacy bare timestamp still guards, and upgrades | `test_a_legacy_bare_timestamp_still_guards_and_is_upgraded` | " |

Deploy note, in addition to the `dedup:34236:*` one from my last review: nothing
migrates `notify_subs_ts`. Existing bare-integer values keep working and upgrade
on the next applied list.

Leaving this as `CHANGES_REQUESTED`, on the pagination half of @dcadenas's review
rather than on anything I pushed. E and F are yours whenever you get to them.

mbradley (commented):
Re-reviewed at `60f195a`, pushed four commits, CI green on `dbfc864`. A is implemented correctly: I mutation-checked the tie-break both directions plus the legacy and exact-replay paths. My `CHANGES_REQUESTED` now rests on the first item below instead.

## The reverse index cannot be repaired once it drifts

If `notify_subs:{subscriber}` is lost while the watcher entries survive, a user who unbells a creator keeps getting pushed about them, permanently, and republishing their list cannot fix it.

`notify_subs` is the only record of which `notify_watchers:*` keys hold that subscriber, and removals are computed from it (`redis_store.rs:398-410`). Lose it and `previous` is empty forever, so the SREM loop is a no-op. This needs no operator error: `DEL KEYS[1]` runs at :407 before the SADD loop at :409, and Redis Lua is atomic but does not roll back, so a script dying in between lands exactly there.

Applying the set difference instead of DEL-then-re-add would leave `notify_subs` a superset of reality on partial failure, which self-heals on the next list, rather than empty, which never does.

## The startup replay is not a rebuild mechanism

If Redis loses the index after a list was applied, those users' bells go silently dead until they happen to re-save their list. Nothing logs it.

This narrows what my `fcb1ef5` bought. It does recover its intended case: a handler failing before the script runs leaves `notify_subs_ts` un-advanced, so the replay re-applies. It does not recover index loss. Once `notify_subs_ts` is advanced, the replay re-fetches the same replaceable event and the guard rejects it on exact-id match. I confirmed against a live Redis: replay returns `false`, watcher set comes back empty. `notify_subs_ts` also has no TTL (`redis_store.rs:415`), so it is a second per-subscriber claim with the same failure shape as the event claim I removed, except that one expired.

@dcadenas asked for "bounded pagination or an equivalent recovery strategy". This is the part pagination does not close. Is the replay meant to be a rebuild mechanism? If yes, the guard needs an escape hatch, such as accepting an exact-id replay when `notify_subs` is missing. If no, write down the recovery procedure, because deleting `notify_subs_ts` is currently the only lever and nothing says so.

## Pushed

**Post-send bookkeeping no longer cascades.** A Redis error after a delivered send used to abort the rest of the function, so a user could get pushed twice about the same video, and tokens FCM had already rejected stayed registered. Both writes now log and continue.

This does not make the two claim records atomic. First-succeeds-second-fails still leaves mention-written, newPost-missing. A pipeline would not close that, since Redis does not roll back on runtime errors. Validating both TTLs before either write would.

**Two tests for properties that had none.** My `170b00b` claimed the watcher-lookup degradation, but its test calls the pure helpers and never reaches `video_notification_targets:966` where it is wired in. At base, no test reached that function at all. The new test fails on exactly the mutation the old one sleeps through. Separately, the rate-limit comment asks that check-then-set not be "fixed" into `SET NX EX`, which now has a test behind it rather than a comment.

**Corrected the event-copy timeout attribution.** `b5edabd` blames the cache-miss cost on `query_timeout_secs`, which no code reads. The bound on that `fetch_events` call is a hard-coded five seconds in `MentionParser::fetch_from_relays`, which is why the measurement was 5.009s and not the configured 3s. The finding was right, the attribution was not, and it matters because an operator debugging that stall would have turned a knob wired to nothing. The warning now sits on the field itself.

One operational note: a bookkeeping failure no longer emits the caller's `"Failed to send notification"`. Both replacements are `error!`, so level-based alerting is fine, but a string-keyed alert goes quiet for this class.

## @realmeylisdev

**`new_post_rate_limit_secs: 0` turns a delivered push into an error.** Reachable without a `settings.yaml` typo: I proved the mapping by loading config, and `NOSTR_PUSH__SERVICE__NEW_POST_RATE_LIMIT_SECS=0` deserialises to `0` with `Settings::new()` returning `Ok` and nothing validating it. `video_coordinate_dedup_ttl_secs` has the identical exposure. Validate at load, or treat 0 as no-throttle and skip the write?

**`a10a02b` only holds on one replica.** An original and its edit have different event ids, so `try_claim_event` does not serialise them. Two replicas handling them at the same moment both find no coordinate record and both send, so a watcher gets the mention and the bell for one video. Either add a sentence to `satisfied_video_claims` saying the guarantee is per-replica, or switch the coordinate claim to `SET NX EX`, which closes it but means a failed send burns the record and that recipient never gets the push. I would take the sentence.

**`b5edabd` caches the fallback, not just the copy.** If the first recipient's profile lookup times out, every remaining watcher in that fan-out sees a short npub instead of the creator's name, where on `main` each recipient retried. Worth a sentence in the `LazyEventCopy` comment, which currently argues only the performance side.

## @rabble

The description says the two missing tests are impossible because "this repo forbids mocks". `MockFcmSender` is on `main`, `fcm_sender.rs` uses it in its own tests there, and it carries `set_error_for_token`. If that is a rule going forward rather than a description, it is not written down and `a10a02b` already depends on it. Which is it? I used it above and will rework those tests if the rule stands.

Ingest authorization and the `d="notify"` namespace are still without a disposition after four reviews.

## Not actioned

- `event_handler.rs:1293` returns `Ok(())` from inside the claim loop on a missing d-tag, skipping invalid-token removal. Same class as the cascade above, but unreachable today because the pre-send gate already returns, so I could not write a failing test and left it.
- `test_replay_horizon_edge_case` is flaky and predates this PR: it builds an event at exactly `now() - 7d` while `is_event_too_old` computes its own, so a second boundary between the two flips the comparison. It will fail in CI eventually and be blamed on whichever PR is unlucky. One line of slack fixes it, unrelated to this change.
- @NotThatKindOfDrLiz, #36 targets this branch and was cut at `4a5a7a8`, before every review commit here. It rewrites `send_notification_to_user` and still has `.await?` on the rate-limit write. Worth rebasing onto current head and keeping the log-and-continue, so the fix is not silently reverted when the fan-out work lands.

## Validation

135 passed / 0 failed locally against Redis on 6379, CI green on `dbfc864`, clippy `-D warnings` and `fmt --check` clean. Each new test verified to fail when its property is removed; each commit reverts independently.

For anyone verifying locally: only the `tests/` suites print a skip line when Redis is absent. The `src/` unit tests skip silently, so a green local run there proves nothing unless `REDIS_URL` points somewhere live. CI provisions `redis:7`.

Leaving this as a comment rather than a verdict. The PR is already blocked, and the pagination decision is @rabble's and @dcadenas's.

realmeylisdev (changes requested):
Takeover, six commits, one per finding so each reverts alone. Gates re-checked
at `dbfc864`: org member with write access, in-repo branch rather than a fork,
non-draft, `divine-push-service` unmapped so the default rule applies. Changed
paths are `src/`, `tests/` and `docs/`. No workflows, no CI, no deployment
manifests.

## The reverse index is fixed, but the trigger you named is not live

`cde41bd`. The `DEL` is gone and the script applies the set difference instead,
so removals clear the reverse index before the forward one and additions write
the forward index first. That holds the invariant at every intermediate step:
every `notify_watchers:*` naming a subscriber has its creator in that
subscriber's forward set. A half-applied run now leaves `notify_subs` a superset
rather than empty.

I reproduced your failure before fixing it, aborting the real script mid-add-loop
by parking a string on one `notify_watchers:*` key so `SADD` raises WRONGTYPE at
a chosen point. On the `DEL` version the forward set comes back `{C}` while
`notify_watchers:A` still names the subscriber, and the unbell that follows is a
no-op — A pushes forever. On the fix it comes back `{A, C}` and the unbell
converges. That is the new test, and it fails on the old script at the superset
assertion.

One correction, in the direction of this being less urgent than written up. Redis
Lua aborts mid-write when a write is *rejected*, and the reachable rejection is
`maxmemory` — which the shared Redis does not set. `k8s/redis-clusters/` carries
no `maxmemory`, no `maxmemory-policy` and no config map, so it runs the defaults:
unlimited, `noeviction`. The memory failure that deployment actually has is the
256Mi container limit, and a SIGKILL does not leave a durable partial script
effect. This was latent, not live.

I kept the fix regardless — it is behaviour-identical on success and free — but
it is worth seeing which way the config line that would make it matter actually
cuts, because it does not cut toward this fix. Under an evicting policy the skew
arrives by eviction rather than by a script aborting: `notify_subs` is one key,
the watcher entries are many, and LRU drops either side independently. The set
difference does nothing about that. Which puts it in your second finding, not
this one.

## The replay as a rebuild mechanism

Not implemented — it is @rabble's and @dcadenas's call — but I checked both halves
against a live Redis so the decision rests on measurements rather than on either
of our readings.

A **total** loss rebuilds. `notify_subs_ts` goes with everything else, the guard
sees no stored value, every replayed list re-applies. Flushed keyspace, replay
returns `1`, watcher set repopulated.

A **partial** loss does not, exactly as you found. Index gone, `notify_subs_ts`
surviving: the replay re-fetches the same replaceable event and the guard rejects
it on exact-id match. Returns `0`, watcher set stays empty.

So the question is narrower than "is the replay a rebuild mechanism". It is:
can this deployment lose the index without losing `notify_subs_ts`? Today, no —
`noeviction` plus a 1Gi PVC makes the loss modes all-or-nothing, and
all-or-nothing recovers. The day someone sets an evicting policy the answer
flips, and it flips for the eviction reason above rather than the script-abort
one.

`da3441e` writes into the developer guide what the replay does and does not
recover, including that deleting `notify_subs_ts` is currently the only lever,
since nothing said so. I deliberately did not write a recovery *procedure* —
choosing one is the design call still outstanding.

## The three you put to me

**`new_post_rate_limit_secs: 0`.** You were right that it is reachable without a
`settings.yaml` typo and I was wrong to write it off as needing one. It is worse
than the two you named: `processed_event_ttl_secs=0` also loads clean, and that
one makes `try_claim_event` error on every event, so nothing is delivered at all.

Of your two options I took validate-at-load (`0222dbf`). Treating 0 as
no-throttle reads the value as an intent nobody holds, turns a protection off
quietly, and would have broken the failure injection in your own two tests, which
need the write to fail. Validation fails at the one moment an operator is
looking, and leaves those tests alone because both zero the field after
`Settings::new()`.

Scope note, since it goes past what you asked: I validated all six numeric bounds
of this shape rather than the two. Three are this branch's;
`processed_event_ttl_secs`, `video_coordinate_dedup_ttl_secs` and
`profile_cache_ttl_secs` predate it and sit in the same one-screen function.
Validating three and leaving three identical holes open seemed the more arbitrary
choice. Drop those lines if you want it narrower. I checked every overlay in
`divine-iac-coreconfig` first — none sets any of the six, so this cannot
crash-loop a running deploy.

**`a10a02b` only holds on one replica.** Agreed, and I took the sentence
(`ef92f6a`). `satisfied_video_claims` now states the guarantee is per-replica,
that what holds it on one replica is `run` awaiting `route_event` inline, and
what `SET NX EX` would cost — which is the price the rate-limit comment already
declines to pay, so the two now read consistently instead of arguing opposite
ways about the same tradeoff.

**`b5edabd` caches the fallback.** Also agreed, also a sentence, same commit. The
`LazyEventCopy` doc now says the cell caches the short-npub fallback along with
the copy, that `main` retried per recipient, and that this is the deliberate
trade rather than an oversight: per-recipient retries are what made one
unreachable profile relay cost five seconds per recipient, and a bell fan-out is
exactly where that multiplies.

## Also pushed

**`726bd1e`** — the `return Ok(())` at the top of the claim loop. Right about the
class; I made it `break`, which is the correct exit regardless since the d-tag
does not vary with `satisfied`. Unreachable today, which is why there is no test:
the pre-send gate already returns under the identical condition. It is my code
from `a10a02b`, and the loop is where that invariant-at-a-distance would break
under a later edit.

**`3852345`** — `test_replay_horizon_edge_case`. Confirmed flaky rather than
assumed: holding one event built at exactly the boundary and re-checking it a
second later flips `is_event_too_old` from `false` to `true`. It now straddles
the cutoff by a minute on each side, which also pins more than the original did,
since the old test only asserted the accept side.

## The mocks question

Factually you are right and the description is wrong. `MockFcmSender` is on
`main`, `fcm_sender.rs` uses it there, and it carries `set_error_for_token`. I
used it too — the config validation had to keep your two injection tests working,
and those are built on it. Whether it is a rule going forward is still @rabble's
to say, but nothing in the repo currently says so.

## A correction to your local-verification note

Your note has the `tests/` suites printing a skip line and the `src/` unit tests
skipping silently. There is a third category: with `REDIS_URL` pointed somewhere
dead, four tests hard-fail rather than skipping — three in `preferences_test`
plus `test_token_reregistration_by_different_pubkey`. Pre-existing and untouched
here, but it means a red local run is not necessarily the reviewer's doing.

The silent-skip warning is worth taking seriously, so rather than trusting the
count I proved the `src/` Redis tests ran: with `test_redis_pool` temporarily
changed to panic instead of returning `None`, all 90 lib tests still pass, so
every one of them got a live pool.

## Still open, unchanged

- Bounded pagination or an equivalent recovery strategy — @dcadenas's other half,
  still the outstanding blocker.
- E (being mentioned costs a watcher the bell they subscribed to) and F (an edit
  re-announces a video the rate limit suppressed). Both reproduced, both
  @rabble's.
- Ingest has no authorization and no bound on distinct subscribers.
- `d="notify"` is a shared user-facing namespace.

## Validation

138 passed / 0 failed against Redis on 6379, up from 135, with no test on the
skip path. `cargo clippy --all-targets --all-features -- -D warnings` clean,
`cargo fmt --check` clean. CI green on `da3441e`.

Each new test checked to fail when its subject is broken:

| Behaviour | Test | Goes red when |
|---|---|---|
| A half-applied list still converges | `test_a_half_applied_list_still_converges_on_the_next_one` | the set difference reverts to `DEL`-and-rebuild |
| Every degenerate zero is rejected | `test_every_degenerate_zero_is_rejected_at_load` | `validate` covers fewer than all six fields |
| The shipped config still boots | `test_shipped_config_passes_validation` | validation is tightened past what `settings.yaml` holds |

@rabble — six commits, each revertible alone. The one wanting your judgement
rather than your review is the config scope note above. Leaving this as
`CHANGES_REQUESTED` on the pagination half, which is where it already was.

NotThatKindOfDrLiz (approved):
Approved after takeover at 48d9d7a. The blocking review items I verified are addressed: paged notify-list replay, legacy video-coordinate dedup reads, rate-limited coordinate recording, muted-mention bell fallback, and idempotent exact replay repair. CI is green. Do not merge until rabble has reviewed the takeover commit and is ready.

realmeylisdev (approved):
Takeover, four commits, `48d9d7a` → `ca072e4`, one per finding so each reverts
alone. Gates re-checked at `48d9d7a`: org member with write access, in-repo
branch rather than a fork, non-draft, `divine-push-service` unmapped so the
default rule applies. Changed paths are `src/`, `tests/` and `docs/`. No
workflows, no CI, no deployment manifests.

## The blocker I left this on is closed

Both halves of it, and I checked the half I was most suspicious of against the
live relay rather than reading it off the diff.

**Pagination.** `48d9d7a` pages backward with `until` and terminates on
`count < limit`. That terminator is only honest if the relay returns the limit
you ask for — a relay that caps below `notify_list_history_limit` would end the
loop on page one while believing it had everything, which would have left the
blocker open under a fix that looked like it closed it. It does not:
`wss://relay.divine.video` returns exactly 5000 events before EOSE for
`limit: 5000`, so the terminator means here what it says.

Two things follow from that rather than being defects to fix now:

- `count < limit` is the only exit in the loop with no log line. If a relay ever
  starts capping, the symptom is silence. A `debug!` there would put the
  difference in the one place it would matter.
- `until` is set to the minimum `created_at` across the merged result. That is
  right for one relay and wrong for two of different depth — the minimum
  advances the query past events the shallower relay never returned.
  `nostr_client` holds only `relay_url`; `profile_relays` live on the separate
  `profile_client`. So this is correct today, and worth remembering before a
  second relay is ever added to that client.

**Recovery.** The `>=` → `>` change closes the case I measured as broken last
round: index gone, `notify_subs_ts` surviving, replay rejected on exact-id
match. It now re-applies and re-asserts the reverse entries.
`test_replaying_the_same_event_repairs_the_reverse_index` pins it.

E and F are both fixed, and reproduce as fixed. A muted video mention falls back
to the bell the watcher enabled; a rate-limited bell records the coordinate so a
later edit stays quiet.

## What I pushed

| Commit | Change |
|---|---|
| `55be9e1` | The bell fallback asks Redis for membership instead of downloading the watcher set |
| `70878cb` | Dead branch out of the notify-list script |
| `8520e08` | The claim recorder can no longer be `?`-ed into skipping invalid-token removal |
| `ca072e4` | The replay-recovery doc says what the code now does |

**`55be9e1`.** The fallback added in `48d9d7a` asks one question — does this
recipient watch this creator — and answered it with `get_notify_watchers`, which
transfers a creator's entire watcher set and parses every member into a
`PublicKey` to keep one bit. That set is the fan-out read, the largest single
value this service reads on a creator with a following, and the fallback ran it
per muted-mention recipient, on top of the copy `video_notification_targets` had
already fetched for the same event moments earlier. `is_notify_watcher` is
`SISMEMBER`. Unparseable members also stop mattering, because the question is
about one member that is known-good. The test asserts the equivalence against
`get_notify_watchers(..).contains(..)` for a watcher, a stranger and a creator
with no watcher key at all.

**`70878cb`.** `had_previous` gated the add loop into two arms that do the same
thing. The `else` skips `SADD KEYS[1] creator`, but only runs when that creator
is already on the forward set, where the `SADD` is a no-op — and the removal
loop cannot have taken it off, because it only removes creators absent from
`incoming` while the add loop iterates `incoming`. Behaviour-identical, so no
test changes; the existing repair and shrink tests already pin both arms. The
cost was the comment, which hangs the exact-replay repair off the `else` and so
reads as though collapsing the branch would take the repair with it — when the
reverse `SADD` that does the repairing is in both arms.

**`8520e08`.** `record_video_claims` logs every write error and returns `Ok(())`
on all paths, but is typed `Result<()>` and both call sites end in `?`. Nothing
propagates today, so this is latent rather than live — the same shape as the
`return Ok(())` that `726bd1e` turned into a `break`. It is worth closing
because of where it sits: the claim writes are the second-to-last thing
`send_notification_to_user` does and invalid-token removal is the last, so one
`?` on `set_cached_string` instead of the `if let Err` converts a bookkeeping
failure into a token FCM has already rejected and nothing retries. Returning
`()` makes that `?` stop compiling.

**`ca072e4`.** `da3441e`'s paragraph on what the replay recovers survived
`48d9d7a` and now says the opposite of the paragraph eight lines below it: that
the guard "rejects it on exact-id match, leaving those bells dead", and that
deleting `notify_subs_ts` is "currently the only lever". Both were true when
written. While correcting it I checked the mirror case against Redis instead of
asserting it — with `notify_subs` gone and the reverse index surviving, an
unbell republished afterwards leaves the dropped creator's watcher entry in
place, because the removal loop has nothing to diff against. The replay still
cannot repair that one, and it is now stated rather than implied.

## One I did not fix, because the measurement says not to yet

An edit of a video published before the subscriber belled the creator is
announced as "posted a new vine". The coordinate record only suppresses what has
a record, and a video from before the subscription has none.

The mechanism to fix it is there: all 299 distinct kind 34236 events I sampled
from the relay carry `published_at`, and nothing in this service reads it. But
the same sample says the symptom is not occurring — `created_at - published_at`
exceeds an hour on one event of the 299 and never exceeds a day, so old vines
are not being edited at a rate anyone would notice. Choosing a threshold for how
old stops being new is a product call, and I would rather not make it against a
problem the data does not show. Worth knowing the tag is there if it ever does.

## Still open, unchanged

- Ingest has no authorization and no bound on distinct subscribers: any pubkey
  can add itself to any creator's `notify_watchers`, which is the hot read path.
  Your call whether that is acceptable while the feature is dark, but it would
  be better tracked as an issue than carried as an open review round.
- `d="notify"` is a shared user-facing namespace.

## Validation

143 passed / 0 failed against Redis on 6379, up from 141, with no test on the
skip path. `cargo clippy --all-targets --all-features -- -D warnings` clean,
`cargo fmt --check` clean. CI green on `ca072e4`: `test` and
`semantic-pull-request` both pass.

Each new test checked to fail when its subject is broken:

| Behaviour | Test | Goes red when |
|---|---|---|
| Membership matches the full read | `test_membership_agrees_with_the_full_watcher_read` | the `SISMEMBER` arguments are transposed |
| A failed claim write still removes a dead token | `test_a_failed_claim_write_still_removes_the_invalid_token` | `record_video_claims` goes back to being fallible and `?`-ed |

@rabble — four commits, each revertible alone. Approving: the pagination blocker
I was holding this on is closed, and I have validated what I pushed. The two
items above are still yours and @dcadenas's; I would rather they became issues
than a reason to keep this open. Merge when you are happy with the four.

mbradley (approved):
Takeover, six commits, `f444836` to `22a1936`, one per finding so each reverts
alone. Gates re-checked at `f444836`: org member with write access, in-repo
branch rather than a fork, non-draft, `divine-push-service` unmapped so the
default rule applies. Changed paths are `src/`, `tests/` and `docs/`. No
workflows, no CI, no deployment manifests.

Your four are answered at the bottom. First the commit nobody had reviewed.

## `f444836` landed after both approvals, and it disabled the safety net

@realmeylisdev and @NotThatKindOfDrLiz approved at `ca072e4`. `f444836` is 481
lines rewriting the video delivery path, merged after that, so it arrived on a
branch whose reviewers had already signed off. It is also the commit I flagged
last round as cut from `4a5a7a8`, before every review commit here.

The good news first: it did not revert anything. I checked all twenty properties
this branch's earlier commits established, and every one holds at head. The
`0bf0c3f` log-and-continue I was most worried about is intact, and I confirmed it
rather than read it, by swapping the rate-limit write back to `.await?` and
watching `test_a_bookkeeping_failure_does_not_skip_the_video_claim` go red.

The problem is what happened to the tests. Moving the delivery rules out of
`merge_watcher_targets` into `watcher_page_targets` and the new paging loop left
six tests pointing at a function production no longer calls. Three mutations,
each leaving the full suite green at `f444836`:

- stop the watcher loop after the first page: 146 passed
- delete the "mention wins on overlap" filter: 146 passed
- make `watcher_page_targets` return nothing at all, so no bell is ever
  delivered: 146 passed

The third is the shape of it. The feature this branch exists to ship could be
disabled outright and CI stays green.

Worth quoting back, because the branch already knew this: `e1c978c`'s own commit
message says the `watchers_or_degrade` tests "never reach
`video_notification_targets`, where the degradation is actually wired in, so
replacing that call with `lookup.expect(...)` leaves the whole suite green."
`f444836` recreated that exact condition, one layer up.

`test_watcher_pages_walk_to_full_coverage` does not cover it either. It seeds
three watchers, and a set below Redis's `set-max-listpack-entries` (128 by
default) is listpack-encoded, so SSCAN returns every member in one reply whatever
COUNT asks for. I measured it: three members with COUNT 2 is one page, two
hundred members with COUNT 2 is eighty-nine. The test walks exactly one page and
passes against an implementation that ignores its cursor entirely.

## What I pushed

| Commit | Change |
|---|---|
| `0e8b764` | The event copy resolves once per event again, not once per page |
| `0de39f7` | Tests for the live bell path, including a real second page |
| `8585d03` | The three `#[cfg(test)]` helpers the rewrite left behind, removed |
| `24076ff` | The watcher-page walk actually crosses a page |
| `f8d205e` | Finding D written down as accepted |
| `22a1936` | A cancelled bell page drains instead of dropping in flight |

**`0e8b764`.** `LazyEventCopy::for_targets` moved inside the page loop, so the
recipient-independent copy is rebuilt for every page. `get_display_name` misses
are never negatively cached and `fetch_from_relays` has a hard-coded five-second
timeout, so a creator with no kind-0 metadata pays a fresh relay round-trip per
page, on the event-handler loop everyone else's notifications queue behind. I
measured it with a counter on `resolve_event_scoped_copy` and 200 watchers:
`page_size=1` gives 152 resolves, `page_size=1000` gives 1. Strictly per page.

Three comments already claimed the property the code had stopped holding, so
hoisting the cell restores what all three describe rather than needing them
rewritten. `needs_content` still comes from the mention targets alone, because
`renders_event_content(NewPost)` is false, so a pure bell fan-out still skips the
content parse.

**`22a1936`.** `send_notifications_bounded` returned at the first `Cancelled`,
which drops the `buffer_unordered` stream and every delivery still running inside
it, at whatever await point each had reached. Up to `concurrency - 1` of them can
be sitting between the FCM send and the writes that follow: the rate-limit
window, the coordinate claim, invalid-token removal. Those pushes have already
shipped, so losing that bookkeeping double-pushes on restart and keeps tokens FCM
has already rejected. It is the same failure `0bf0c3f` made the post-send writes
log-and-continue to avoid, arriving through a different door, and it is silent,
because a dropped future logs nothing where the `tokio::select!` arms were
written to log exactly this. Sequential delivery could strand one recipient;
a bounded page strands fifty.

Scope note, since I would rather state it than have it found: this does not close
the window inside `send_notification_to_user`, which still returns `Cancelled`
between the FCM send and the writes below it. That one predates this branch.

Two one-line fixes ride along in that commit because neither has behaviour of its
own to revert separately. The mid-walk page-failure log said "delivering mentions
without remaining bells", which reads as though no bells shipped when a failure on
page seven has already delivered six; it now names the abandonment and carries the
delivered count. And `get_notify_watchers`' doc still called it "the read for the
fan-out" when it has no caller in `src/` at all: it is the unbounded `SMEMBERS`
this commit's parent exists to keep off the hot path, surviving as a test helper.

**`8585d03`.** `merge_watcher_targets`, `watchers_or_degrade` and
`split_delivery_targets` are all `#[cfg(test)]` at head. The last one production
never called in any revision; its only caller is a test asserting a partition of a
literal it just built, which no production change can turn red. The cost was not
the dead lines, it was that the "mention wins on overlap" rationale sat on the
dead function while `watcher_page_targets` carried no doc at all, and three
comments elsewhere still sent readers to the dead one for the rule.

## Your four

**A.** Fixed, and it survives everything that landed after it. I re-verified at
head rather than trusting the earlier round: inverting the lowest-id comparison in
the Lua turns three tests red.

**B.** Fixed, at `event_handler.rs:93` and `:160`. Agreed the claim was buying
nothing, for the reason you give.

**C.** Fixed, and your framing of it was better than mine. Note it regressed in
`f444836` from per-recipient to per-page, which `0e8b764` closes.

**D.** Your reasoning holds and I have no objection to deferring the code. The
half that was missing was writing it down, so `f8d205e` does that: what is
accepted, why it is safe today (public kind 30000 list, no tokens so nothing
sends), what it actually costs (wasted fan-out iteration), and the trap that makes
it more than a one-liner. No behaviour change. The guide previously read as though
the gap were an oversight.

## Three for you, none of them blocking

**1. `new_post_delivery_concurrency: 50` against `connection_pool_size: 10`.**
Each delivery takes six to eight separate pool acquisitions and bb8's
`connection_timeout` is 15s, so a wide fan-out oversubscribes the pool five to
one. I could not turn that into a failure: 150 recipients at concurrency 50
against a pool of five delivered all 150 with no errors, and an independent run at
200 recipients agreed. So it queues rather than drops on a healthy Redis, and I am
asking rather than claiming. The new `validate()` checks each field is positive
but never the relationship, and `test_new_post_fanout_limits_are_bounded` asserts
`concurrency <= page_size`, which as far as I can tell protects nothing. Raise
`connection_pool_size`, validate `concurrency <= connection_pool_size` at load, or
document and accept?

**2. The PR body is stale against `f444836`.** Rollout Notes lists three settings;
that commit added `new_post_fanout_page_size` and `new_post_delivery_concurrency`,
both defaulted and operator-facing, and both joined the must-be-positive list, so
a zero via `NOSTR_PUSH__SERVICE__*` now aborts startup. That belongs in the
rollout notes. Validation says 93 lib tests, head is 95. And Decisions Worth
Review does not mention that the delivery shape changed: mentions and bells are no
longer one merged target list, mentions go first sequentially and bells page with
bounded concurrency, so "mention wins on overlap" is now an ordering plus a set
difference rather than a merge rule. That last one is squarely a decision worth
review.

**3. `FCM_BATCH_CONCURRENCY = 100` is hard-coded** in a commit whose whole point is
making these bounds configurable and validated, and it changed `send_batch` for
every notification kind rather than just new-post. Probably moot, since a user
holds one to three tokens, but it multiplies with the 50 on paper and it is the
one bound in this area an operator cannot reach. Config, or a comment saying why
it does not need to be?

## Validation

145 passed, 0 failed against a live local Redis, with no test on the skip path.
`cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo fmt
--check` clean. CI green on `22a1936`: `test` and `semantic-pull-request` both
pass.

The count went 146 to 145 because I removed seven tests that asserted on code
production cannot run and added six that assert on code it does. Each new test
checked to fail when its subject breaks:

| Behaviour | Test | Goes red when |
|---|---|---|
| Every watcher is belled, not just page one | `test_the_fan_out_reaches_watchers_past_the_first_page` | the loop stops after one page, or the cursor never advances |
| A mentioned watcher is not also belled | `test_a_mentioned_watcher_is_not_also_belled` | the `mentioned` filter is dropped |
| Watchers are typed as new-post | `test_a_watcher_page_types_every_watcher_as_new_post` | `watcher_page_targets` yields nothing |
| An author watching themselves is dropped | `test_a_watcher_page_drops_an_author_watching_themselves` | the self filter is dropped |
| Partial overlap keeps the non-mentioned watchers | `test_a_watcher_page_keeps_the_watchers_who_were_not_mentioned` | either filter is dropped |
| A cancelled page reports cancellation | `test_a_cancelled_bell_page_reports_cancellation` | the `Cancelled` arm is swallowed |

Two things I could not put behind a test, stated rather than hidden. `0e8b764`'s
property has no assertion, because there is no seam to count relay round-trips
without a mock behind `MentionParserService`; it is verified by reading and by the
counter measurement above. And hardcoding the page size and concurrency past the
settings still leaves the suite green, for the same reason: the delivery count is
identical, only the page count differs.

One operational check, since the widened validation is the part that could bite on
deploy: no overlay in the iac repo sets any of the eight validated fields. The
only `NOSTR_PUSH__SERVICE__*` overrides in the manifests are `PRIVATE_KEY_HEX` and
`ALLOWED_PUBKEYS`, so this cannot crash-loop a running deploy.

## Still open, unchanged

Ingest authorization and the bound on distinct subscribers, and the `d="notify"`
shared namespace. Not restating them, only noting they are still without a
disposition. @realmeylisdev's suggestion that they become issues rather than
another review round seems right to me.

@rabble, six commits, each revertible alone. Approving: my `CHANGES_REQUESTED`
from Jul 31 is what was blocking this, and all four of the findings behind it are
now resolved. The three questions above are yours, and none of them needs to hold
the merge.

PR #36, opened 2026-07-31
fix(push): bound new-post fanout delivery

NotThatKindOfDrLiz:
## Summary
- rebase onto current `feat/new-post-notifications` and keep the base branch's graceful watcher-lookup degradation
- replace the permanent NewPost watcher cap with paged Redis `SSCAN` fan-out, so every bell subscriber can be reached without one unbounded watcher read
- deliver each NewPost watcher page with bounded recipient concurrency, and make per-recipient FCM token sends concurrent inside `send_batch`
- add validation for the new paging/concurrency knobs and regression coverage for full watcher-page coverage and mention delivery when watcher paging fails
- keep Redis-backed tests aligned with the existing skip-when-unavailable convention so local full-suite verification is reliable

## Motivation
PR #35 adds bell notifications where recipient count is audience-shaped rather than p-tag-shaped. The handler needs bounded Redis reads and bounded delivery work, but dropping all subscribers past a fixed cap would silently penalize popular creators and the people who explicitly enabled bells.

This version treats `new_post_fanout_page_size` as a work-size bound, not a product recipient limit: the handler continues paging until Redis returns cursor 0. Mention targets still deliver even if the watcher lookup path fails.

## Linked issue
Refs #33.
Refs #35.
Refs #37.

## Verification
- `cargo test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The service subscribes to trigger events on its relay and notifies the tagged re
 | Repost | 16 | Repost of a user's note (NIP-18) |
 | New post | 34236 | A creator the user subscribed to ("belled") published a video |
 
-New-post notifications are the one type not anchored to a `p` tag on the trigger event. Recipients come from the subscriber's own NIP-51 list (kind 30000, `d=notify`), so the service resolves them from a Redis reverse index rather than from the video. They are rate-limited to one push per (subscriber, creator) per hour; the in-app feed is not throttled. See [the protocol doc](docs/nip-xx-push-notifications.md) for the list shape.
+New-post notifications are the one type not anchored to a `p` tag on the trigger event. Recipients come from the subscriber's own NIP-51 list (kind 30000, `d=notify`), so the service resolves them from a Redis reverse index rather than from the video. They are rate-limited to one push per (subscriber, creator) per hour, and fan-out is paged and delivered with bounded concurrency so one popular creator cannot force one unbounded Redis read or sequential delivery loop. The in-app feed is not throttled. See [the protocol doc](docs/nip-xx-push-notifications.md) for the list shape.
 
 Divine video comments are NIP-22 `kind:1111` and notify both the root video author and the direct parent author (deduplicated when they coincide). Follows (kind 3) are defined in the protocol and subscribed to, but new-follow notifications are **not currently emitted** — that requires diffing contact-list state, which is not yet implemented.
 
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -15,8 +15,10 @@ service:
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
-  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list can't stall Redis
+  notify_list_max_creators: 1000  # Bounds the Lua diff so one huge list cannot stall Redis
   notify_list_history_limit: 5000  # Page size for historical notify-list replay
+  new_post_fanout_page_size: 1000  # Watcher SSCAN page size for one video
+  new_post_delivery_concurrency: 50  # Concurrent NewPost deliveries per page
 
 redis:
   url: "redis://127.0.0.1:6379"
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -15,8 +15,10 @@ service:
   processed_event_ttl_secs: 604800  # 7 days
   video_coordinate_dedup_ttl_secs: 31536000  # 1 year
   new_post_rate_limit_secs: 3600  # One new-post push per (subscriber, creator) per hour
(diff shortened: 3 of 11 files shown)


rabble (commented):
The sequential/bounded split is clean, config and docs move together with the code, and the tests are real rather than decorative. `futures-util` is already a dependency and `handle_content_event` is awaited rather than spawned, so the 50 is a true global ceiling on in-flight sends — that all checks out.

The bounded-concurrency half of this should merge. My concern is entirely with the watcher cap.

## The cap drops bell subscribers, and probably the same ones every time

`new_post_fanout_max_watchers: 1000` means a creator with more than 1000 belled subscribers notifies 1000 of them per video. `SSCAN` iterates hash buckets in reverse-binary cursor order, so starting from cursor 0 visits the same buckets first for a given table size. For a subscriber set that is not being resized, an early-terminated scan returns substantially the same members every time.

So it is not "a random 1000 each video". It is closer to a stable subset who get notified and a stable remainder who never do — silently, with no signal to them, having explicitly tapped the bell.

The number also has no product meaning. `new_post_rate_limit_secs` is a real product limit: it protects people from a prolific creator training them to disable notifications, and there is a user-facing rationale for it. `new_post_fanout_max_watchers` protects the event-handler loop, and charges subscribers for it. Those are different categories and I do not think they should both sit in the config as tuning knobs.

The practical impact today is probably nil — during invite-only rollout few creators are likely over 1000. It starts costing exactly when a creator gets popular, which is when their notifications matter most and when nobody is reading a `warn!` line.

I would rather we did not ship this as a permanent knob. If it stays as a safety valve for now, it needs a `TODO` pointing at the paging work, because once it is just another config value "why did this creator's fans stop getting notified" becomes a very hard question to answer six months from now.

## The cap bounds accepted watchers, not scanned members

```rust
for member in members {
    if watchers.len() == limit { truncated = true; break; }
    match PublicKey::from_hex(&member) { ... }
}
```

`watchers.len()` only grows on a parseable, non-duplicate member. If a creator's set contains more than `limit` unparseable entries, this scans the whole set — the unbounded work the PR exists to prevent. The doc comment says the hot path "never materializes an unbounded watcher set", which is true of memory but not of iteration. Worth bounding total members examined, or loop rounds, as well as accepted watchers.

## `Vec::contains` is O(n²) on the hot path

`if !watchers.contains(&pubkey)` is linear, inside the per-member loop: roughly 500k comparisons at the default 1000. Fine at that size, but this is a performance change and the bound is configurable. At 10k it is ~50M comparisons per video, which would undo the point of the PR. A `HashSet` for membership alongside the `Vec` for order keeps it O(n).

## The truncation warning can fire when nothing was truncated

The `watchers.len() == limit` check runs before dedup, so once `limit` distinct watchers are collected, the next member flips `truncated` even if it is a duplicate of one already held. `SSCAN` can return duplicates across iterations, so a set of exactly `limit` distinct members can log "remaining watchers skipped" when none were. Given that warning is the only evidence for the problem in the first section, false positives are worth avoiding.

## Minor

- `test_new_post_fanout_limits_are_bounded` asserts exact values (`1000`, `50`), so tuning the YAML breaks CI. The relationship assertions below are the durable part; the equality assertions may be deliberate, but they are the brittle bit.
- The bounded path has no `token.is_cancelled()` pre-check like the sequential one, so up to `concurrency` deliveries can start after cancellation. Small.
- `get_notify_watchers` is now unused in `src/` and referenced only by tests. Worth a doc comment saying it is not for the hot path, so it does not get reintroduced.
- The Redis `PING` guards in `preferences_test.rs` and `test_token_reregistration.rs` are unrelated to fan-out. Justified in the description, just noting the mixed scope.

## The reason I think the cap is avoidable

Filed separately as #37: `FcmClient::send_batch` is a sequential `for` loop of `send_single` awaits, so each recipient's devices are notified in series inside the new 50-way concurrency. That is where the real throughput ceiling is, not in the watcher lookup.

Fixing that and raising the concurrency ceiling gets a 100k-subscriber creator notified in roughly ten seconds with every per-user control intact and nobody dropped. The durable version is to page `SSCAN` with a cursor and run fan-out off the event-handler loop entirely — enqueue `(event_id, creator, cursor)`, deliver a page, re-enqueue the next cursor until the cursor returns to 0. A complete `SSCAN` iteration returns every element present throughout, so full coverage comes for free once we stop terminating early. `notify_rate:{subscriber}:{creator}` already acts as a dedup key for resumed jobs, and `try_claim_event` already handles replica safety.

None of that needs to happen in this PR. But it is the reason I would keep the cap explicitly temporary rather than settle on 1000 as the answer.

rabble (approved):
Re-reviewed current head b0d9f94. The permanent watcher cap from my prior review is gone; SSCAN pagination reaches full coverage, per-page delivery remains bounded, FCM token sends are bounded internally, and zero-valued concurrency/page settings fail at startup. Local validation with Redis on 6379: cargo fmt --check, cargo clippy --all-targets --all-features -- -D warnings, and cargo test (146 tests plus doc test) all pass. Ready for the author to merge.

PR #38, opened 2026-08-05
docs: archive engagement notification brief

rabble:
## Summary

- replace the duplicated engagement-notification discovery draft with a short pointer to the canonical plan
- record the proposed pull-based delivery direction and explicitly reject the superseded inbound campaign API
- link the canonical defect tracker instead of duplicating its findings here

## Motivation

The full discovery draft had already drifted from the canonical `divine-engagement` copy and described an inbound push-service API that is no longer the proposed direction. A small pointer keeps this repository searchable without creating another plan copy that can go stale.

The pointer distinguishes current behavior from the proposal: push-service currently exposes `/health` and `/metrics`; the poller, consent enforcement, quiet hours, and frequency caps described by the proposed delivery contract are not implemented yet.

## Linked issue

Refs divinevideo/divine-engagement#2. The issue remains open because the canonical plan still needs correction.

## Manual test plan

- run `git diff --check`
- verify the current routes against `src/server.rs` and `README.md`
- verify the proposed endpoint direction and unimplemented controls against `divine-engagement/docs/push-service-contract.md`
- verify the duplicated-plan defects link to divinevideo/divine-engagement#2


docs/plans/marketing-engagement-campaign-platform.md
```diff
@@ -0,0 +1,1015 @@
+# Plan: marketing and engagement campaign platform
```
NotThatKindOfDrLiz, line 1: The canonical plan in divine-engagement says this discovery draft should link there rather than carry its own copy. Keeping a 1,000-line copy here creates immediate doc drift in the repo most likely to implement stale push-service guidance. Please replace this body with a short archive pointer to the canonical plan and the current push-service contract.

docs/plans/marketing-engagement-campaign-platform.md
```diff
@@ -0,0 +1,1015 @@
...
+actor_identity
+action
+reason
+metadata_json
+occurred_at
+```
+
+Audit events are append-only through application code.
+
+## Internal push-service contract
```
NotThatKindOfDrLiz, line 509: This contract is stale and points implementation in the wrong direction. The current divine-engagement push-service contract says push-service has no ingress and should poll GET /api/internal/deliveries/pending, then POST /api/internal/deliveries/results back to divine-engagement. Leaving POST /internal/v1/campaigns/* in this repo risks someone building the rejected ingress/API path.

.gitignore
```diff
@@ -24,6 +24,9 @@ dump.rdb
 test_results.txt
 dm_event.json
 prompt-plan.md
+/tmp/
+/output/.DS_Store
+/.worktrees/
```
NotThatKindOfDrLiz, line 29: This rule is redundant: the existing .DS_Store rule already ignores output/.DS_Store at any depth. The neighboring /tmp/ and /.worktrees/ additions are also unrelated local-environment changes in a docs archive PR, so please drop these ignore edits unless the PR introduces a repo producer for them.

NotThatKindOfDrLiz (changes requested):
Requesting changes. This PR should not merge as-is because it archives a superseded cross-service plan in the push-service repo, documents the rejected inbound push-service API direction, and tracks an unreproducible PDF artifact under a scratch output path. Please replace the long plan body with a short pointer to the canonical divine-engagement plan and push-service contract, remove or regenerate/relocate the PDF with source, and drop the unrelated ignore rules.

mbradley (commented):
Agreed with Liz's direction, not restating it.

Five defects in the plan body that are also present in the canonical `divine-engagement` copy, so they need fixing there whatever happens to this PR:

**1. The expiry example teaches the opposite of the expiry rule.** Lines 527/537 and 564/575 pair an `expiresAt` of `12:00:00Z` with a `notBefore`/`retryAfter` of `19:00:00Z`, a deferral seven hours past expiry. Lines 607-608 say an expired message is dropped rather than delivered late.

**2. The frequency caps have no precedence rule.** Lines 617-623: the per-category maxima sum to roughly 5 per 7 days against a combined non-social cap of 2, and product education alone (2 during onboarding) consumes the entire budget. One sentence on which cap yields would settle it.

**3. FunnelCake is assigned ownership it does not have.** Lines 27 and 327-342 make it the source of behavioural and engagement facts. FunnelCake is the Nostr backend: relay plus REST over ClickHouse. It stores events, not app-engagement data. Is a different source intended here, or is this new work for FunnelCake?

**4. The unknown-timezone policy is both decided and open.** Line 611 states suppress as the default and line 182 lists it as an MVP exclusion, while lines 941 and 984 still carry it as an open question.

**5. Unit mismatch.** Line 425 stores `holdout_basis_points`; line 740 requires "a holdout percentage".

mbradley (approved):
Took a look at `0667a15`. My five points are handled. They belong to the canonical copy, not this one, and you've filed them as `divinevideo/divine-engagement#2` (including the holdout basis-points-vs-percent mismatch, which is a real 100x risk if built as written). Liz's three are addressed here too — the long body is now a pointer, the direction is stated correctly, and the PDF and ignore-rule edits are gone. **Approving.**

**Keep or close (your question):** keep it. A short, findable note in this repo saying "don't build the incoming-API version, this service polls instead" is worth more than a README line — this is the repo where someone would otherwise build the wrong thing, and the pointer says so plainly.

NotThatKindOfDrLiz (approved):
Takeover fixes are verified on `8afcd72`. The pointer now distinguishes current routes from the proposed poll-based contract, states that the poller and policy controls are not implemented, and links the canonical tracker instead of duplicating its findings. The full diff is clean and CI is green.

PR #42, opened 2026-08-21
fix(push): stop two recoverable conditions from permanently halting all delivery

realmeylisdev:
Two recoverable failure modes could stop push delivery while the process and health endpoint remained up. This change keeps the relay listener alive after broadcast lag and contains FCM send panics to the affected token.

## 1. Continue after a lagged broadcast receive

`notifications().recv()` returns a bounded Tokio broadcast receiver. `RecvError::Lagged(n)` means the subscriber fell behind, but receiving can continue from the oldest retained message. Only `Closed` is terminal.

The listener now warns with the dropped count and continues after `Lagged`. It still stops on `Closed`, shutdown, or cancellation.

## 2. Contain FCM send panics

`send_batch` runs inside the event-handler task. A panic while sending one token must not unwind through that task and stop later notification processing.

Each token send is now wrapped at the batch boundary. A panic becomes an `InternalResponse` error for that token while sibling sends continue. Token-prefix logging also truncates on character boundaries, including error and invalid-token cleanup paths, so malformed strings cannot escape the containment through logging.

## Tests

- The live listener loop is driven over an intentionally overflowed broadcast channel, then checked for forwarding retained notifications after `Lagged`.
- Listener tests also cover `Closed` and cancellation exits.
- A panicking FCM mock uses a multibyte token and verifies the batch returns all three results while both healthy tokens succeed. The test enables error tracing so the panic fallback log path is exercised.

## Verification

- `cargo test --verbose`

## Not included

Additional unrelated notification-delivery follow-ups remain outside this change.


mbradley (approved):
Approving. Both fixes are right, and I confirmed both failure modes are real: the listener really did quit for good on a brief bit of back-pressure, and one panicking send really did take down the whole handler task — with `/health` staying green through both. Good finds, and the writeup made them easy to check.

One thing worth a look before merge, on the lag fix:

The lag test checks the small helper that decides "lagged → keep going, closed → stop," but nothing actually runs the loop that had the bug. So if someone later changed that loop back to quitting on a lag, this test would still pass — the exact regression it's named for would slip through. The panic test doesn't have that gap: it runs a real send through the batch and shows the other recipients still go out.

I don't think it blocks merge — the fix itself is correct. But the loop is the part that broke, so it's the part worth pinning down. Is testing it directly awkward enough (needing a live relay feed) that checking the helper is the sensible stopping point, or is there a clean way to reach the loop's error path? Your call — just flagging it so it's deliberate rather than by accident.

realmeylisdev:
Answered by testing it — the loop was reachable, so it's tested now: ec24c57.

`process_live_events`'s only tie to `self` was the one line fetching the receiver, and `notifications()` hands back a plain `broadcast::Receiver<RelayPoolNotification>` whose variants are public. No live relay needed. Extracting the loop as `run_live_loop(receiver, event_tx, service_pubkey, token)` lets a test drive the real thing over a `broadcast::channel` overflowed on purpose: four notifications into a two-slot channel, so the loop's first `recv()` is `Lagged(2)`, then assert the two retained behind it are still forwarded.

I checked it actually catches the regression rather than merely passing — putting `break` back on the `Lagged` arm fails the test with "loop stopped on the lag instead of continuing".

Since the loop is directly testable now, `recv_action`/`RecvAction` was a seam with nothing left to do. It's folded back into the loop's match arms, rationale kept as a comment, and both public items are gone — one fewer name in the crate's API than before this PR. That also picks up two exits the seam never reached: `Closed` still stops the loop, so "continue on any error" isn't a valid way to fix the lag bug, and cancellation still ends it.

The extraction itself is a pure move. I diffed the old and new loop bodies normalized for whitespace and comments; the only difference is the `recv_action` call becoming the two `RecvError` arms.

149 tests pass, clippy and fmt clean.

dcadenas (approved):
The recoverable failure paths are contained now.

### Changes I pushed

- Made every FCM token-prefix log character-safe.
- Exercised panic fallback logging with a multibyte regression token.

### Checks

- `cargo test --verbose`

The Semantic PR reusable workflow has no local command; its GitHub check was green at publish time.

Issue #43, opened 2026-08-21
fix(fcm): dead tokens are never pruned, and live tokens expire 90 days after registration


realmeylisdev:
Two independent defects in the same subsystem — token lifecycle in Redis. Together they mean the token set drifts away from reality in both directions: dead tokens accumulate, live tokens get deleted.

## 1. `UNREGISTERED` tokens are never pruned

`FcmError::from` classifies a dead token by substring-matching the FCM error body:

```rust
// src/fcm_sender.rs:69-76
if error_message.contains("invalid registration token")
    || error_message.contains("registration token is not a valid FCM registration token")
    || error_message.contains("BadDeviceToken")
    || error_message.to_lowercase().contains("unregistered")
    || error_message.to_lowercase().contains("not registered")
{
    FcmError::TokenNotRegistered
```

But that body only ever arrives for HTTP **400**. The pinned client discards it for every other 4xx:

```rust
// firebase-messaging-rs @ f2cb78b0, src/lib.rs:213-220
StatusCode::BAD_REQUEST => {
    let data = hyper::body::to_bytes(res.body_mut()).await…;
    Err(E::from(RPCError::InvalidRequest { details: data }))
}
e if e.is_client_error() => Err(E::from(RPCError::invalid_request())),   // <- body dropped
```

`InvalidRequest` with no details maps to a constant string (`src/fcm_sender.rs:89-91`), which matches none of the substrings above.

| FCM condition | HTTP | Body reaches us? | Pruned? |
|---|---|---|---|
| `INVALID_ARGUMENT` | 400 | yes | yes |
| `UNREGISTERED` | 404 | **no** | **no** |
| `SENDER_ID_MISMATCH` | 403 | **no** | **no** |
| `QUOTA_EXCEEDED` | 429 | **no** | n/a |

`UNREGISTERED` is the normal signal for an uninstalled or reinstalled app, so this is the common case. The `SENDER_ID_MISMATCH` branch we wrote (`:73-78`) is unreachable for the same reason.

**Effect:** a dead token stays in `user_tokens:{pubkey}` and costs an extra sequential FCM round-trip on every notification to that user, until the 90-day sweep. That directly worsens the throughput problem in #37.

Classifying on the HTTP status rather than the body text would fix it, but the status is not surfaced either — this likely needs an upstream change or a thin wrapper.

## 2. A live token is deleted 90 days after its last *registration*, not its last use

The staleness score is written **only** in `add_or_update_token`:

```rust
[...]

PR #47, opened 2026-08-21
fix(push): replace abandoned FCM client, supervise tasks, and report delivery liveness in /health

dcadenas:
## Summary
- Push delivery stopped completely for about 59 hours while both pods reported `2/2 Running` with zero restarts.
- Any FCM 5xx panicked the delivery task, and nothing noticed: the panic was swallowed, shutdown was never triggered, and `/health` returned 200 the whole time.
- This change removes the dependency that panicked, bounds every outbound call, and makes a dead delivery pipeline visible to Kubernetes instead of invisible.

## Motivation
- The last delivered notification was at 2026-08-19T01:00:24Z. One pod had been silently dead since 2026-08-12.
- `firebase-messaging-rs` read the retry hint on every FCM 5xx via `HeaderName::from_static("Retry-After")`. `http` rejects uppercase header names by panicking, so that line panics every time it runs. Upstream has not shipped since October 2024 and still has the bug.
- The panic killed the event-handler task, which closed the event channel and ended the listener. `main` discarded the `JoinError`, so `token.cancel()` never ran and the process stayed up serving a health check that reported only its own pubkey.
- Recovery required someone to notice and restart the pod by hand.

## Related Issue
- Closes the first half of #43. `UNREGISTERED` (404) and `SENDER_ID_MISMATCH` (403) now reach the classifier, because the response body is read on every status rather than only on 400. The 90 day expiry described in the second half of #43 is unchanged and still open.
- Together with #42 this clears the first blocker listed in divinevideo/divine-iac-coreconfig#1871.
- Related to #37. `send_batch` is already concurrent on `main`, so only the second half of that issue (fan-out running on the event-handler loop) still applies.

## Testing
- `cargo clippy --all-targets --all-features` is clean and `cargo test` passes 184 tests, including the tests added by #42.
- New regression tests cover the failure that caused the outage: a 503 carrying `Retry-After` now returns `RetryableInternal` instead of panicking, and the header is read case-insensitively.
- End to end against a local relay and Redis, with the service in Docker. A NIP-44 encrypted kind 3079 registration was stored, and a kind 7 reaction targeting that pubkey resolved the recipient, checked preferences, found the token and reached the FCM send. With a stub credential the send fails as `Unauthorized`, and the token is left in Redis rather than deleted.
- Also verified in Docker: the image builds, the binary links no OpenSSL, `/health` returns 200 when healthy, a dead task drives a non-zero exit, and `SIGINT` still exits 0.
- Not tested: real delivery to a device. That needs production credentials, so the FCM boundary is as far as local verification goes.

## Protocol / Ops Notes
- `/health` now returns 503 when the listener or event handler is dead, and adds a `tasks` object. The 200 body keeps its existing shape. Both probes already point at `/health`, so no manifest change is needed.
- A task that ends unexpectedly now cancels the others and the process exits non-zero, so Kubernetes restarts the pod instead of keeping a dead one in service.
- On restart the live subscription replays only the last hour, so a restart drops very little.
- Token pruning now requires positive evidence that the token is dead: an `UNREGISTERED` or `NotRegistered` error code, or a message naming a registration token. A bare 404 no longer deletes a token. Deleting on status alone would erase live registrations in bulk during a routing or project misconfiguration.
- Outbound calls are bounded: 30s per FCM request, 10s to connect, and a 45s ceiling on the whole send. The ceiling matters because credential fetching happens before any HTTP request exists and the auth library sets no timeout of its own.
- The relay connect path now waits for the connection instead of checking immediately after a call that only starts one in the background. That check raced the handshake and lost on cold start.
- `firebase-messaging-rs`, `gcloud-sdk`, `tonic`, `http 0.2` and all of OpenSSL leave the dependency tree. `gcp_auth` replaces them for credentials and resolves the same Application Default Credentials the service already used, so no credential or deployment change is needed. `libssl` is dropped from the image.
- The delivery summary log now reports failures that are not dead tokens. It previously counted only pruned tokens, so an outage where every send failed still logged `failed_count=0`.

- Confirmed. No partner, customer, brand or campaign names appear in the title, branch name, description or code. No credentials, tokens or user pubkeys are included.


hm21 (approved):
Approved. Good change — the panic, the swallowed `JoinError`, and the always-200 `/health` were three separate reasons that outage stayed invisible, and this closes all three. The dead-token classification requiring positive evidence rather than a bare 404 is the right call.

@dcadenas I pushed four commits to your branch; please re-review the head. Each is independent and revertable on its own.

**Pushed**

- `f8329d9` — the failure branch of the send summary switched to `error = ?result.as_ref().err()`, which logs `error=Some(Unauthorized("..."))`. Restored `Display`, so the field reads as the reason again. Log text only.
- `0be8737` — `tests/fcm_panic_resilience_test.rs`'s module doc still said, in the present tense, that `main` never supervises the event handler and that `firebase-messaging-rs` panics on every 5xx. Both are what this branch fixes. Reworded to past tense and stated what the test still guards.
- `ada42b2` — `/health` gained a `tasks` object and a 503 path, but `README.md` and `docs/developer-guide.md` still documented the old always-200 shape; the repo guide asks for doc updates on operational behaviour changes. Documented both responses, noted `pubkey` is present either way, and added one sentence on supervision to the architecture paragraph.
- `b8e927c` — **not requested by the review, flagging it explicitly.** The shutdown contract this PR builds is gated on `ctrl_c()`, which is SIGINT only. Kubernetes stops a pod with SIGTERM, and with no handler the default disposition kills the process outright — so task cancellation, the HTTP drain, and the new non-zero exit never run in the one environment they were written for. The verification bullet in the description says "`SIGINT` still exits 0", which is true and never happens in production. Now waits on both. Handler-install failure falls back to the old SIGINT-only behaviour instead of failing startup; non-unix targets unchanged; worst case is still the kubelet's SIGKILL at `terminationGracePeriodSeconds`. I confirmed both signals resolve the shutdown future in a binary running the helper. `shutdown_signal` sits in the binary crate, so there is no unit-test seam for it — say the word if you'd rather have it reverted.

**Left for you**

1. `gcp_auth::provider()` in `RealFcmClient::new` is the one gcp_auth call this PR does not bound, and the reasoning for `FCM_OPERATION_TIMEOUT` applies to it verbatim: it runs inside `AppState::new`, before the HTTP server binds, against the same metadata server. Not an outage — I checked the manifest and `livenessProbe` (`initialDelaySeconds: 10`, `periodSeconds: 10`, default `failureThreshold: 3`) reaps a hung startup after roughly 40s. But it dies by SIGKILL with nothing in the log, where a `timeout` would name the cause. Your call whether it is worth the extra failure mode.
2. The Docker image is never built on a pull request — the `docker` job is `if: github.event_name != 'pull_request'`. This PR replaces the Dockerfile's runtime deps and the whole dependency tree, so the first CI validation of the image happens on `main`. Worth knowing: the highest declared MSRV in the resolved tree is now exactly `1.85` (`gcp_auth 0.12.7`, `base64ct 1.8.1`, `backon 1.6.0`) against the `rust:1.85.1` builder, so there is no headroom left — the next routine `cargo update` can break the image build with every PR still green. I did not touch the workflow: that is a platform-sensitive path and I'm on `@divinevideo/reviewers`, not `@divinevideo/platform`.
3. Nit, no action needed: `classify_error` maps 429 to `InvalidRequest`. A quota burst is retryable rather than invalid, and it reads oddly in the logs. The message body still carries the truth, and nothing retries today, so it only matters if `RetryableInternal` ever grows a consumer.

**Validation** — on the pushed head, locally: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --lib` (133 passed). CI is green on `b8e927c`: `test`, `semantic-pull-request`, `license/cla`. Redis-backed integration tests and the Docker build ran in CI only; I have neither locally. Not verified by anyone yet: real delivery to a device, as your description already says.

@mbradley heads-up, the head moved since your review request.

mbradley (approved):
Approving.

A single server-side error from Firebase crashed the one task that sends notifications, and nothing caught it. The pod reported healthy with zero restarts while delivering nothing for 59 hours.

The fix addresses all three reasons it stayed invisible:

- The library carrying the crash is gone, replaced by a direct call to Firebase's current API using the same credentials the service already had, so nothing new needs configuring or deploying.
- A dead delivery task now takes the process down instead of lingering, and health reports unhealthy the moment either half of the pipeline stops, so Kubernetes restarts the pod rather than trusting a green check.
- Every outbound call is time-bounded, including the credential fetch that runs before any request exists, so a peer that goes quiet can't park delivery again.

Token deletion got the closest look, since a wrong delete silently and permanently stops a real person's notifications. It's tightened correctly: a token is removed only on positive proof the device dropped it, and the loose text matching that could have wiped out live tokens during an infrastructure mix-up is gone, with tests covering those false-alarm cases.

The panic catch that keeps a delivery crash from taking down the handler stays. The specific bug it was written for is gone, but it still protects any sender swapped in later.

Tests and lint are clean, and it builds against the new dependencies.

PR #50, opened 2026-08-22
feat(observability): expose push delivery metrics

dcadenas:
## Summary

Add application metrics that show whether the push pipeline is receiving events, completing routing attempts, and delivering through FCM. The existing health endpoint still reports task liveness; the new metrics make a live but inactive or failing delivery path observable.

Refs #48.

## What Changed

- Added `GET /metrics` on the existing HTTP port with `Cache-Control: no-store`.
- Added relay-event, event-processing, FCM attempt/success/failure, and token-pruning counters with bounded reason labels.
- Added the last-processed-event timestamp used by the event-processing deadman.
- Added endpoint and metric tests, including mixed FCM outcomes and confirmed token-pruning counts.
- Documented metric semantics and how the deadman complements FCM failure and task-health alerting.

## Testing

- `cargo test --verbose`

The reusable Semantic PR workflow has no local command and was not run locally. I did not test a live VictoriaMetrics scrape or alert delivery; those require the companion infrastructure change.

## Follow-up

The infrastructure repository still needs the service scrape and alert rules described in issue #48.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,6 +479,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-epoch"
+version = "0.9.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
+name = "crossbeam-utils"
+version = "0.8.22"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "61803da095bee82a81bb1a452ecc25d3b2f1416d1897eb86430c6159ef717c17"
+
+[[package]]
 name = "crunchy"
 version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -576,6 +591,8 @@ dependencies = [
  "futures-util",
  "gcp_auth",
  "hex",
+ "metrics",
+ "metrics-exporter-prometheus",
  "nostr-relay-builder",
  "nostr-sdk",
  "once_cell",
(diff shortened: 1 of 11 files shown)


realmeylisdev (approved):
Checked this against the `divine-funnelcake` precedent #48 points at, and against how the service is actually deployed. It does what the issue asked, and the instrumentation sits where it can't quietly lie.

Verified locally on `a7269bf` with Redis up:

- 187 tests green, `clippy -D warnings` and `fmt --check` clean. Ran the suite 12x, plus `--test-threads=1` and `--test-threads=32` — no flakiness.
- Drove the real startup path (`init()` → global recorder → `render()`), which the tests only reach through local recorders. Valid exposition output, `# HELP`/`# TYPE` present, and the startup gauge comes out a genuine unix timestamp rather than a zero.
- Nothing enters the pipeline uncounted: all three `event_tx.send` sites are preceded by `event_received()`, and `send_batch` is the only production send path, so no delivery escapes the FCM counters.
- Cardinality is closed. `metric_reason()` is an exhaustive match over fixed literals and drops the error payload, so FCM response text can't ride into a scrape as a label. Ceiling is about 16 series.
- `/metrics` is on the existing internal service port with no ingress, so it isn't publicly reachable.

One thing that looks wrong and isn't, in case it comes up in review: `install_recorder()` never runs upkeep in production. In this crate version `run_upkeep()` only drains histograms, and this PR registers none, so counters and gauges render live either way.

Three notes, none blocking:

**For the IaC follow-up rather than this PR** — the production rollout allowlist is still in place, so the FCM counters will read near zero until it lifts. A plain "no successful sends" rule would fire on that alone today, well before it could catch a real outage.

**Worth a follow-up** — when the relay notification receiver lags, `nostr_listener.rs` already has the dropped count in hand and only warns it. Those are silently lost pushes, which is the exact failure shape #48 exists to surface, and it's the one gap these counters don't cover.

**Minor** — `init()` is the only piece here with real logic (registers descriptions, seeds the startup gauge that the deadman's grace window depends on) and nothing tests it. I exercised it directly and it behaves, so this is a missing regression guard rather than a live gap. A small `tests/` binary would cover it, since `set_global_recorder` only fires once per process.

PR #51, opened 2026-08-22
fix(fcm): deliver Android social notifications promptly

### LINKED ISSUE #44: fix(fcm): Android notifications are sent at normal priority and get deferred by Doze
Android messages are built data-only with no `android` block at all:

```rust
// src/fcm_sender.rs:211-220
let message = Message::Token {
    token: token.to_string(),
    name: None,
    notification,
    data: payload.data,
    android: None,   // <- no AndroidConfig, so no priority
    apns,
    webpush: None,
    fcm_options: None,
};
```

With no `AndroidConfig.priority`, FCM applies its default of **normal** priority for data-only messages. Normal-priority messages are batched and withheld while the device is in Doze or App Standby.

**Effect:** on Android a like/comment/repost notification can be delayed by minutes to hours on an idle device — exactly the state a phone is in when a push is the only thing that would bring the user back.

iOS is unaffected: `build_apns_config` sets `apns-priority: 10` and `apns-push-type: alert` explicitly (`src/fcm_sender.rs:179-183`).

The data-only choice for Android is deliberate and documented (`docs/developer-guide.md:65`) so the client controls rendering — that part is fine. Priority is simply not set alongside it, and the doc does not mention priority at all.

## Fix

Set `AndroidConfig { priority: High, .. }` on the message. High priority is the correct classification here: these are user-visible, user-triggered social notifications, which is what FCM's high priority is for.

Worth pairing with a note in `docs/developer-guide.md` so the data-only + high-priority combination is recorded as intentional.

## Confidence

Verified by reading source at `ed8b410`. The Doze/priority behaviour is Firebase-documented, not repo-verified.

Found during an end-to-end delivery audit; see `divinevideo/divine-iac-coreconfig#1871` for context.

dcadenas:
## Summary

Android social notifications can be delayed while a device is idle because Firebase Cloud Messaging defaults them to normal priority.
This change requests high-priority Android delivery while preserving the data-only payload that lets the app render one notification.

## What Changed

- Added the supported high-priority Android configuration to Firebase HTTP v1 messages.
- Kept the Android payload data-only with no top-level notification field.
- Added wire-level coverage for the serialized request and documented the intentional platform contract.

## Testing

- `cargo test --verbose`
- Not tested on a physical Android device.

The reusable Semantic PR workflow has no local command and will run after the pull request is opened.

Closes #44

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -64,7 +64,7 @@ The service watches for these event kinds and notifies the tagged recipient:
 
 The FCM message carries **no top-level `notification` field** — the `data` map below is always present and is identical in shape for every notification type (only the `title`/`body` strings differ); every `data` value is a string. Per-platform delivery then diverges so that **one incoming push produces exactly one visible banner**:
 
-- **Android** — data-only (`notification` and `android` unset). Android does not auto-display data messages, so the app renders the single banner itself from the `data` fields.
+- **Android** — data-only (top-level `notification` unset) with `android.priority` set to `high`. Android does not auto-display data messages, so the app renders the single banner itself from the `data` fields; high priority lets FCM wake an idle device promptly for this user-visible notification.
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Takeover review complete. Android high priority is now limited to payloads with visible notification copy, silent/background payloads retain default priority, and wire coverage protects both paths plus the existing notification branch. Local formatting, clippy, and full tests pass; required CI is green.

PR #52, opened 2026-08-22
fix(push): make recipient delivery retry-safe and fan-out durable

### LINKED ISSUE #45: fix(dedup): a per-event claim plus no retry loses notifications permanently
Two defects that combine: the dedup claim is taken before delivery and is never released, and nothing retries. Any failure after the claim loses that notification for seven days.

## 1. The claim is per-event, but the spec says per-recipient

`docs/nip-xx-push-notifications.md:113` specifies:

> At most one notification per (recipient_pubkey, app, event_id)

The implementation claims a single key for the **whole event**, before any recipient is processed:

```rust
// src/event_handler.rs:99-126
claim_result = redis_store::try_claim_event(
    &state.redis_pool, &event_id,
    state.settings.service.processed_event_ttl_secs,
) => { … }
…
let handler_result = route_event(&state, &event, context, token.clone()).await;
```

`try_claim_event` is a plain `SET NX EX` with no release on error (`src/redis_store.rs:264-277`).

Recipients are then fanned out sequentially (`src/event_handler.rs:441-472`). An event with five p-tags that dies after recipient 2 — pod eviction, OOM, a panic — loses recipients 3 through 5 permanently: a restart re-reads the event within the 1-hour `since` window, and `try_claim_event` returns `false`.

Making the key `(event_id, recipient_pubkey)` as the spec already says would make partial fan-out recoverable.

## 2. Nothing retries

`grep -rni "retry\|backoff" src/` finds only the `RetryableInternal` variant declaration and its `From` arm (`src/fcm_sender.rs:32-33`, `:88-90`). It is constructed and then discarded:

```rust
// src/event_handler.rs:715-721
Err(e) => {
    error!(
        target_pubkey = %target_pubkey, token_prefix = truncated_token, error = %e,
        "FCM send failed for token"
    );
}
```

So a transient FCM failure — quota, 5xx, a network blip — drops the notification, and the 7-day claim guarantees it is never reprocessed. The error type that exists specifically to say "retry this" is the one being thrown away.

Other paths that abandon a recipient with the claim already held: a Redis error in `get_tokens_for_pubkey` (`:602-604`) or `get_user_preferences` (`:617-622`), both via `?`.

## Fix sketch

- Key the claim on `(event_id, recipient_pubkey)` per the spec, so one recipient's failure cannot consume another's.
- Honour `RetryableInternal` with a bounded backoff, or release the claim on a retryable failure so the next delivery of the same event can pick it up.

## Confidence

Verified by reading source at `ed8b410`, including the grep for retry handling. The spec/code divergence is quoted from the repo's own `docs/nip-xx-push-notifications.md`.

Related: #42 fixes the panic that makes the crash-mid-fan-out case reachable today. Found during an end-to-end delivery audit; see `divinevideo/divine-iac-coreconfig#1871`.

dcadenas:
## Summary

A failure for one notification recipient could suppress every recipient still waiting on the same event.
Large new-post audiences also kept the main event handler busy until every watcher page finished.
This change makes recipient failures independently recoverable and moves watcher delivery into durable background work.

## What Changed

- Claim content delivery per event and recipient immediately before sending.
- Release claims after confirmed retryable failures while retaining successful or ambiguous deliveries.
- Queue new-post watcher pages in Redis with leases, bounded concurrency, exponential retry backoff, and expiry.
- Supervise the new worker through the existing health and shutdown lifecycle.
- Preserve mention ordering, preferences, rate limits, video-edit deduplication, and invalid-token cleanup.

## Testing

- `cargo test --verbose`

Redis-backed tests covered partial-recipient replay, retryable claim release, lease recovery, immediate graceful rescheduling, atomic page advancement, retry-state persistence, and multi-page delivery.
The reusable Semantic PR workflow has no local command and will run only on GitHub.

## Risks

Watcher pages use at-least-once processing.
Completed recipient claims prevent page retries from resending successful deliveries, while lease expiry recovers a worker crash.

Issue #37's concurrency tuning remains separate because it depends on live FCM project quota evidence.

Closes #45
Addresses #37

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,8 +72,11 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 - `token_to_pubkey` - Hash mapping a token back to its owner
 - `stale_tokens` - Sorted set of token timestamps, for cleanup
 - `user_preferences:{pubkey}` - User's notification preferences (JSON)
-- `dedup:{event_id}` - Per-event processing claim with TTL. Notify lists are exempt: they are idempotent by `created_at`, so the claim prevents nothing and a failed handler would otherwise strand the list for the TTL
+- `dedup:{event_id}` - Per-event processing claim for control events
+- `dedup:{event_id}:{recipient}` - Per-recipient content-delivery claim. Retryable failures release it; successful or ambiguous sends retain it
 - `dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify. Scoped by notification type so a bell does not suppress a later mention on the same video. The scoping is one-directional: a delivered mention also writes the `newPost` record, because naming the video already tells the recipient it exists
+- `fanout:enqueued:{event_id}` - Initial durable new-post fan-out enqueue marker
+- `new_post_fanout_jobs` - Sorted set of durable cursor-page jobs, scored by availability or lease expiry
 - `notify_subs:{subscriber}` - Creators this user has belled
 - `notify_subs_ts:{subscriber}` - `created_at:event_id` of the last applied notify list (out-of-order guard). The id resolves a `created_at` tie by NIP-01's lowest-id rule; a bare integer from an earlier build still reads as a timestamp with no known id
 - `notify_watchers:{creator}` - Subscribers watching this creator (hot read path)
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ The metrics endpoint exposes:
 | `push_fcm_sends_attempted_total` | Counter | FCM sends attempted per device token. |
 | `push_fcm_sends_succeeded_total` | Counter | Successful FCM sends per device token. |
 | `push_fcm_sends_failed_total{reason}` | Counter | Failed FCM sends by bounded failure reason. |
-| `push_new_post_fanout_retries_total{reason,outcome}` | Counter | Durable new-post page retries by bounded failure reason and scheduled/exhausted outcome. |
+| `push_new_post_fanout_retries_total{reason,outcome}` | Counter | Durable new-post page retries by bounded failure reason and scheduled, preserved, or exhausted outcome. |
 | `push_tokens_pruned_total{reason}` | Counter | Tokens removed as `invalid` or `stale`. |
 | `push_last_event_processed_timestamp_seconds` | Gauge | Unix timestamp of the last completed event routing attempt. Initialized at startup to give a new instance one alert window. |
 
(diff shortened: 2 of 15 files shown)


mbradley (approved):
Reviewed for delivery correctness, the durable queue's concurrency, trust & safety, and tests. Queue design now prevents retries from double-notifying and losing watchers, and the rate limit holds. I pushed one fix, have one question, and one note.

**Pushed (`ef16e36`):** FCM 429 ("slow down") responses were treated as permanent, so during a large fan-out the affected watchers were never retried — the case this change exists to survive. Now a 429 is retried, honoring the wait FCM asks for. All delivery tests pass; I confirmed they fail without the fix.

**Question — should a network timeout be retried?**
> *As someone who's turned on new-post notifications for a creator, when the push to me times out on the way to Google and isn't retried, I may never get told about their new video — even though nothing else was wrong.*

A 429 is safe to retry (nothing was sent). A timeout is ambiguous — the push might have gone through, so retrying risks sending it twice. Right now timeouts aren't retried: an occasional miss is traded to avoid an occasional duplicate. Timeouts are the most common transient failure, so this decides which is the better failure — "missed it" or "got it twice." If not-retrying is the intended trade-off, a one-line comment locks it in.

**Note — brief duplicate pushes during a deploy, which I think is fine to accept.**
> *As anyone using the app during a deploy, I get the same like/comment/mention/repost notification twice for a few minutes.*

The "already sent this" bookkeeping changed shape, so while the old and new versions run side-by-side (every deploy, and any rollback) they don't recognize each other's markers and those notifications can double up briefly. My read is this is an acceptable trade: the window is minutes and only during a deploy, it heals itself once fully rolled out, video "new post" alerts aren't affected, and a duplicate is a minor annoyance rather than a missed, wrong, or leaked notification. The alternative — having the two versions share the old marker during the overlap — adds complexity and would partly reintroduce the failure this PR removes. Suggest just adding a line to the PR description noting it (applies to rollback too). Flag it if you'd rather coordinate.

mbradley (approved):
Re-approving on the latest commit. The changes since my last review hold up, and I ran the tests and checks locally to confirm they pass.

What this fixes, in plain terms:

- **One person's failed notification no longer blocks everyone else waiting on the same post.** Each recipient now succeeds or retries on its own.
- **Notifying a large creator's followers no longer ties up the service** while it works through the whole audience. That work now runs in the background and recovers on its own if a worker goes down.
- **Retries are safe now:** nobody gets notified twice, and nobody gets silently dropped. A post that keeps failing to send gives up after a set number of tries instead of retrying for days, and the followers queued behind it keep getting served.

Solid improvement over where we were. Thanks for the careful work on the retry handling.

Optional, non-blocking follow-up: when the service does give up on part of an audience, the logs and metrics don't capture which post or how many followers were affected. Worth adding so we'd notice if anyone ever silently misses a notification.

PR #53, opened 2026-08-22
fix(push): stop subscribing to unsupported follow lists

### LINKED ISSUE #46: chore(notifications): kind 3 is subscribed and firehosed but has no handler
Kind 3 (NIP-02 follow list) is subscribed, streamed, claimed in Redis, and then discarded. No follow notification is ever produced.

```rust
// src/event_handler.rs:406-411
} else if kind_num == 3 {
    // Kind 3: Contact list - notify newly followed users
    // Note: This would require tracking previous contact list state
    // For now, we skip this as it requires state comparison
    debug!(event_id = %event_id, "Contact list event - follow notifications not yet implemented");
    return Ok(());
```

The README is accurate about this (`README.md:41`), so it is a known gap rather than a surprise. What is worth fixing is the cost of leaving it wired up:

- **Bandwidth.** Kind 3 is in `notification.event_kinds` (`config/settings.yaml:39`) and the subscription is a firehose — kinds plus `since`, no `#p`, no `authors`, no `limit` (`src/nostr_listener.rs:189-191`). Every contact-list update on the relay crosses the wire. Follow lists are among the largest events on Nostr, and NIP-02 requires a full republish of the entire list on every single follow, so this is the heaviest kind per byte.
- **Redis.** Each one takes a 7-day claim in the same keyspace as real notifications (`src/event_handler.rs:99-126`) before being dropped.
- **Throughput.** The handler is fully sequential, so this traffic contends with real deliveries — see #37.
- **A dead user-facing toggle.** Kind 3 is in `default_preferences.kinds` (`config/settings.yaml:49`) and in the spec's own preferences example, so users are offered a follow-notification setting that is on by default and does nothing. `NotificationType::Follow` is unreachable dead code.

## Options

Either implement it or unwire it — the current middle state pays the cost without the feature.

Implementing it needs previous-contact-list state to diff against, since a naive p-tag notification would fire "X followed you" on every unrelated edit to the list, forever. That is real work and a real product decision.

Unwiring it is small: drop `3` from `notification.event_kinds` and from `default_preferences.kinds`, and remove the `NotificationType::Follow` arm. Worth doing now if follow notifications are not imminent, and reverting when they are.

Same question applies to kind 30023 (`config/settings.yaml:43`): Divine does not publish long-form articles, so that subscription exists only to catch third-party posts that happen to p-tag a Divine user, on a tag NIP-23 does not define and NIP-27 makes optional. Also a firehose, also low yield.

## Confidence

Verified by reading source and config at `ed8b410`. Found during an end-to-end delivery audit; see `divinevideo/divine-iac-coreconfig#1871`.

dcadenas:
## Summary

The service subscribes to every kind 3 contact-list update even though it cannot identify new follows and never sends a follow notification.
This change removes that unused traffic and the dead follow preference and payload path.

## What Changed

- Removed kind 3 from relay subscription and default preference lists.
- Removed unreachable follow notification handling and copy.
- Updated service docs and tests while keeping generic protocol examples service-neutral.

## Testing

- `cargo test --verbose`

Three Redis-backed preference tests skipped locally because Redis was unavailable.
CI provisions Redis and will exercise them.
The Semantic PR reusable workflow has no local command and will run after the PR opens.

## Risks

Existing kind 3083 preferences can retain a kind 3 entry, but it is inert because no notification type maps to it.
Follow delivery remains out of scope and is tracked in [divinevideo/divine-funnelcake#204](https://github.com/divinevideo/divine-funnelcake/issues/204).
Kind 30023 behavior is unchanged.

Closes #46

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,7 +84,6 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 |------|------|-------------|
 | Like | 7 | Reactions to user's notes |
 | Comment | 1 | Replies to user's notes |
-| Follow | 3 | New followers |
 | Mention | 1 | Notes mentioning user |
 | Repost | 16 | Reposts of user's notes |
 | NewPost | 34236 | A belled creator published a video (recipients from `notify_watchers`, not `p` tags) |
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The service subscribes to trigger events on its relay and notifies the tagged re
 
 New-post notifications are the one type not anchored to a `p` tag on the trigger event. Recipients come from the subscriber's own NIP-51 list (kind 30000, `d=notify`), so the service resolves them from a Redis reverse index rather than from the video. They are rate-limited to one push per (subscriber, creator) per hour, and fan-out is paged and delivered with bounded concurrency so one popular creator cannot force one unbounded Redis read or sequential delivery loop. The in-app feed is not throttled. See [the protocol doc](docs/nip-xx-push-notifications.md) for the list shape.
 
-Divine video comments are NIP-22 `kind:1111` and notify both the root video author and the direct parent author (deduplicated when they coincide). Follows (kind 3) are defined in the protocol and subscribed to, but new-follow notifications are **not currently emitted** — that requires diffing contact-list state, which is not yet implemented.
+Divine video comments are NIP-22 `kind:1111` and notify both the root video author and the direct parent author (deduplicated when they coincide). Follow notifications are not handled by this service, so it does not subscribe to kind 3 contact lists.
 
 Each FCM message carries a stable `data` payload with routing and presentation fields. Routing to the correct video uses the authoritative addressable coordinate from the triggering event, never a coordinate synthesized from the recipient's pubkey. The full payload contract is documented in the [developer guide](docs/developer-guide.md#fcm-payload-format).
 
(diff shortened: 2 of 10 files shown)


realmeylisdev (approved):
Approved — checked this all the way down, and it really is dead-code removal rather than a behaviour change.

`NotificationType::Follow` is never constructed anywhere in this repo's history. It only ever appeared in two `match` arms, so `type: "follow"` has never reached a device and no client can break on its removal.

What I ran:

- Both binaries against a local relay with `APP_ENV=production`. `main` sends REQ kinds `[1, 3, 7, 16, 1111, 3079, 3080, 3083, 30023, 34236]`; this branch sends the same list without `3`. The kind 30000 `#d=notify` subscription is untouched.
- The k8s deployment sets `APP_ENV=production` and there is no `NOSTR_PUSH__NOTIFICATION__*` override in divine-iac-coreconfig, so `config/settings.yaml` is what actually runs. The config change ships.
- A stray kind 3 now lands in the existing "no notification handler" arm — same early return as the branch you deleted.
- A saved `3` in someone's kind 3083 preferences stays inert, since `is_kind_enabled` is only ever reached through `NotificationType::kind()`.
- divine-mobile is the only consumer of this payload contract org-wide, and its push router already treats unrecognised types as non-fatal. divine-web has no push path at all.
- 184 tests pass locally against a live Redis, none skipped, plus clippy `-D warnings` and `fmt`. The three preference tests your description mentions do pass once Redis is up.

Pushed one fix: 50cdae1 renames `test_runtime_config_event_kinds_match_defaults`, which now also guards `default_preferences.kinds`. Test-only, no assertion touched.

One note left inline about the client half of #46's dead toggle — not something this PR should carry.

PR #54, opened 2026-08-23
fix(push): stop deregistering devices that are still receiving notifications

### LINKED ISSUE #43: fix(fcm): dead tokens are never pruned, and live tokens expire 90 days after registration

realmeylisdev:
## Summary
- A delivered push now refreshes the device token's score in the `stale_tokens` sorted set, so the 90-day cleanup sweep deletes devices that have gone quiet instead of devices that merely registered a long time ago.
- Only a *delivered* push counts. A failed send leaves the score untouched.
- The refresh is one `ZADD stale_tokens XX GT CH` for the whole batch of a recipient's tokens.

## Motivation

The cleanup sweep deletes a device token 90 days after its score in `stale_tokens`. That score was written in exactly one place — `add_or_update_token`, at registration — and nothing ever moved it afterwards.

So the sweep was measuring time since the device last *registered*, not time since it was last reachable. A device that had been receiving push notifications the entire time was deleted anyway, silently, with no error surfaced on any path. The user just stops getting notifications and has no way to know why.

The devices this hits hardest are the ones that cannot recover on their own. iOS grants background refresh only to apps the user opens regularly, and never to a force-quit app. Android's periodic work is deferrable and gets skipped under Doze. A dormant install cannot re-register itself — and a push notification is the one thing that would bring the user back to the app to trigger a re-registration. The sweep switches that off precisely for the users who need it.

## Why this fix

A delivered push is the only positive evidence this service ever gets that a device still exists. Recording it makes the sweep mean "inactive", which is what the 90-day window was always intended to mean.

Failed sends deliberately do not count. An auth failure or an FCM outage says nothing about whether the token is good, and treating those as activity would keep genuinely dead tokens alive forever — the opposite of what #47 just fixed.

Both `ZADD` flags are load bearing, and each is pinned by a test that fails if the flag is removed:

- **`XX`** never creates a member. A token deregistered or swept between the send and this write must stay gone. Re-adding it would leave a tracked token with no `token_to_pubkey` owner and no membership in any user set — an orphan that nothing clears until it ages out a second full max-age window.
- **`GT`** never lowers a score. Replicas stamp this from their own clocks, so without it a peer running behind could pull a live token *toward* the sweep, which is the exact outcome this change exists to prevent.

## What this deliberately leaves alone

- **A user who receives no notifications at all for 90 days is still swept.** Delivery-based refresh cannot help someone nothing is being delivered to. That residual gap needs provider feedback as the authority, not a timer.
- **The NIP-40 `expiration` tag is still ignored.** `docs/nip-xx-push-notifications.md` requires an `expiration` tag and says servers MUST ignore expired events; the service does neither, and its 90-day sweep agrees with the client's 90-day `expiration` by coincidence rather than by design. That is a spec decision, not an implementation detail, so it is filed separately rather than settled here.
- **No change to the sweep itself**, its inclusive boundary, or `token_max_age_days`.
- **No change to pruning.** Dead-token detection landed in #47 and is untouched.

## Related Issue
- Closes #43. The first half of that issue (dead tokens never pruned) landed in #47; this is the second half.
- The NIP-40 `expiration` spec decision that was tangled into #43 is split out into #55 and is deliberately not addressed here.

## Testing

Local, against Redis 7.4.11 in Docker:

- `cargo test` — 190 passed, 0 failed.
- `cargo clippy --all-targets --all-features` — clean.
- `cargo fmt --check` — clean.

Six new tests. Each was watched failing before the implementation existed:

| Test | Pins |
|---|---|
| `a_refresh_moves_a_registered_tokens_score_forward` | the score actually moves |
| `a_refresh_never_resurrects_a_token_that_is_no_longer_registered` | the `XX` flag |
| `a_refresh_never_moves_a_score_backwards` | the `GT` flag |
| `a_refresh_of_no_tokens_touches_redis_not_at_all` | empty batch is a no-op, not an error |
| `test_a_delivered_push_refreshes_the_tokens_staleness_score` | the wiring, end to end through `send_notification_to_user` |
| `test_a_failed_send_leaves_the_staleness_score_alone` | a failed send is not treated as proof of life |

The two flag tests were verified as real guards by mutation: removing `XX` fails the resurrection test and only that test; removing `GT` fails the backwards test and only that test.

`ZADD ... XX GT` semantics were also checked directly against Redis 7.4.11 before being relied on: `XX` does not create an absent member, `GT` does not lower an existing score, and a multi-member call silently skips the absent ones.

## Protocol / Ops Notes

- **Requires Redis 6.2+** for the `GT` flag. Dev and CI run Redis 7; the cluster manifests pin 8.4. No action needed, but it is a real floor now.
- **No migration, no backfill.** Existing scores stay valid — they are already "last known good", just only ever written at registration. Tokens start being refreshed on the next delivered push.
- **One extra Redis round trip per recipient per delivered notification**, batched into a single command across all of that recipient's tokens.
- New `error!` log `"Failed to refresh token activity after a delivered push; the sweep may deregister a live device"`. It logs and continues rather than propagating: the push has already shipped, so failing here would report a delivered notification as failed and skip the bookkeeping that follows it.
- The `stale_tokens` key description in `AGENTS.md` and `docs/developer-guide.md` is updated, since what the score means has changed.

- Confirmed. No partner, customer, or brand names, and no real user identifiers. Test tokens and pubkeys are generated per-run.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,7 +70,7 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 ### Redis Keys
 - `user_tokens:{pubkey}` - Set of FCM tokens registered for a pubkey
 - `token_to_pubkey` - Hash mapping a token back to its owner
-- `stale_tokens` - Sorted set of token timestamps, for cleanup
+- `stale_tokens` - Sorted set scored by the last time a token was known good, for cleanup. Written at registration and again on every delivered push, so the sweep means "inactive" rather than "registered long ago". The refresh uses `ZADD XX GT`: `XX` so a token deregistered mid-send is not resurrected as an owner-less member, `GT` so a replica's clock cannot pull a live token toward the sweep
 - `user_preferences:{pubkey}` - User's notification preferences (JSON)
 - `dedup:{event_id}` - Per-event processing claim with TTL. Notify lists are exempt: they are idempotent by `created_at`, so the claim prevents nothing and a failed handler would otherwise strand the list for the TTL
 - `dedup:{kind}:{type}:{owner}:{d-tag}:{recipient}` - Per-recipient video delivery, so a NIP-33 edit does not re-notify. Scoped by notification type so a bell does not suppress a later mention on the same video. The scoping is one-directional: a delivered mention also writes the `newPost` record, because naming the video already tells the recipient it exists
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ The service is a single async binary running four cooperating tasks: a Nostr lis
 ### Prerequisites
 
 - Rust 1.85+
-- Redis
+- Redis 6.2+
 - A Firebase project with FCM enabled, and a service-account credentials file
 
 ### Development
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -393,7 +393,7 @@ the fan-out follow-up rather than done by halves here.
 |-------------|------|-------------|
 | `user_tokens:{pubkey}` | Set | FCM tokens registered for a pubkey |
 | `token_to_pubkey` | Hash | Reverse mapping from token to owner pubkey |
-| `stale_tokens` | Sorted Set | Token timestamps for cleanup |
+| `stale_tokens` | Sorted Set | Last time each token was known good, for cleanup. Scored at registration and refreshed on every delivered push, so the sweep deletes devices that have gone quiet rather than devices that merely registered a long time ago. The refresh is `ZADD XX GT`: `XX` never re-creates a token deregistered between the send and the refresh, `GT` never lowers a score |
 | `dedup:{event_id}` | String | Per-event processing claim with TTL. Not taken for notify lists, which are idempotent by `created_at` and would be lost for the TTL if a failed handler left a claim standing |
 | `dedup:34236:{type}:{owner}:{d-tag}:{recipient}` | String | Per-recipient video delivery decision, retained for the configured coordinate TTL (one year by default). `{type}` is the notification type (`newPost`, `mention`), so a bell and a mention for the same video coordinate keep independent records. A delivered mention writes both records, since naming the video already tells the recipient it exists; a delivered or rate-limited bell writes its own |
 | `user_preferences:{pubkey}` | String | JSON notification preferences |
--- a/src/event_handler.rs
+++ b/src/event_handler.rs
@@ -1547,7 +1547,10 @@ async fn send_notification_to_user(
 
     // Process results
     let mut tokens_to_remove = Vec::new();
-    let mut success_count = 0;
+    // Doubles as the success count. A delivered push is the only evidence this
+    // service gets that a device still exists, so the tokens are worth keeping
+    // rather than just counting.
+    let mut delivered_tokens = Vec::new();
     // Counted separately from `tokens_to_remove`: a dead token is a *pruned*
     // token, not the only way a send fails. Reporting only removals meant an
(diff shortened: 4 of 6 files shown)


NotThatKindOfDrLiz (approved):
Approved after takeover verification. The Redis 6.2+ prerequisite is now documented, the full diff remains clean, and the post-push CI test, clippy, and formatting steps all ran and passed. The adjacent cleanup atomicity debt is tracked separately in #58.

PR #57, opened 2026-08-23
fix(relay): recover silent event subscriptions

dcadenas:
## Summary

A relay connection can stay open after its event subscriptions stop delivering, leaving a pod healthy but ineffective.
This change watches event flow, reissues stalled subscriptions, and fails listener health when recovery does not restore delivery.

Closes #56.

## What Changed

- Added a configurable five-minute event-silence timeout. One silent interval reissues both live subscriptions; another fails the supervised listener so the process restarts.
- Gave both subscriptions stable IDs and reacts immediately when the relay closes either owned ID. Closures for both IDs from the same relay-side action share one recovery attempt; a repeated rejection of a recovered subscription fails the listener instead of entering a replay loop.
- Backs off briefly after a rate-limited closure before replaying, then allows a later recovery after healthy event flow has resumed.
- Replays from a one-hour `since` timestamp on startup and recovery so events from the stalled window can be processed. Successful event IDs are retained for seven days in the shipped configuration, preventing duplicate push delivery from the overlap.
- Requires subscription requests to be queued specifically for the configured main relay, checks connectivity for that relay specifically, and raises `CLOSED` and `NOTICE` messages to info logs.
- Resets the silence deadline only after an event reaches the handler queue, so downstream backpressure does not masquerade as relay silence.
- Threads shutdown cancellation through recovery subscriptions.
- Preserved the receive metrics from main and the removal of unsupported kind 3 from the shipped configuration.

## Testing

- `cargo test --verbose`

The reusable Semantic PR workflow has no local command and was not run locally.
I did not test against the production relay or restart any deployed workload.

Regression coverage includes paired owned-subscription closures, repeated recovered-subscription rejection, rate-limit backoff, recovery after healthy traffic resumes, event-flow silence, unrelated closures, downstream handler backpressure, cancellation, and listener health propagation.

## Risks

The two owned subscriptions share a five-second recovery grace period. One closure for the sibling ID during that period is treated as belonging to the same relay-side action; a genuine immediate rejection of only that sibling would instead be detected by the event-silence watchdog if flow does not resume.

The silence timer tracks the combined event stream, so ordinary traffic cannot prove that the low-volume notify-list subscription remains active. An explicit `CLOSED` for that subscription is handled immediately, but silent relay-side loss without a message remains detectable only through broader delivery monitoring.

The one-hour timestamp asks the relay for the stalled window; any relay-side response cap can still limit how much history is returned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Any setting can be overridden with the `NOSTR_PUSH__` prefix and `__` as the nes
 | `NOSTR_PUSH__SERVICE__PRIVATE_KEY_HEX` | Yes | Service's Nostr private key (hex), used for NIP-44 decryption |
 | `NOSTR_PUSH__REDIS__URL` | No | Redis connection URL (overrides the config file) |
 | `NOSTR_PUSH__NOSTR__RELAY_URL` | No | Nostr relay to subscribe to |
-| `NOSTR_PUSH__NOSTR__EVENT_SILENCE_TIMEOUT_SECS` | No | Quiet period before resubscribing, then failing health if silence continues (default `300`) |
+| `NOSTR_PUSH__NOSTR__EVENT_SILENCE_TIMEOUT_SECS` | No | Quiet period before the listener resubscribes, and the window after any resubscribe before it fails health (default `300`) |
 | `APP_ENV` | No | Selects the config file (default `development`) |
 | `RUST_LOG` | No | Log level (default `info`) |
 
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Any setting can be overridden with the `NOSTR_PUSH__` prefix and `__` as the nes
 | `NOSTR_PUSH__SERVICE__PRIVATE_KEY_HEX` | Yes | Service's Nostr private key (hex), used for NIP-44 decryption |
 | `NOSTR_PUSH__REDIS__URL` | No | Redis connection URL (overrides the config file) |
 | `NOSTR_PUSH__NOSTR__RELAY_URL` | No | Nostr relay to subscribe to |
+| `NOSTR_PUSH__NOSTR__EVENT_SILENCE_TIMEOUT_SECS` | No | Quiet period before resubscribing, then failing health if silence continues (default `300`) |
 | `APP_ENV` | No | Selects the config file (default `development`) |
 | `RUST_LOG` | No | Log level (default `info`) |
 
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -2,7 +2,11 @@
 
 nostr:
   relay_url: "wss://relay.divine.video"
-  event_silence_timeout_secs: 300  # Resubscribe after 5 quiet minutes; fail health after 5 more
+  # Quiet window the listener tolerates before acting on relay silence.
+  # A cold silence resubscribes once, then fails health if a second window
+  # passes with no event. A relay CLOSED resubscribes immediately, so there
+  # the next quiet window is the one that fails health.
+  event_silence_timeout_secs: 300
   profile_relays:
     - wss://relay.divine.video
     - wss://purplepag.es
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -2,6 +2,7 @@
 
 nostr:
   relay_url: "wss://relay.divine.video"
+  event_silence_timeout_secs: 300  # Resubscribe after 5 quiet minutes; fail health after 5 more
   profile_relays:
     - wss://relay.divine.video
     - wss://purplepag.es
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -2,7 +2,11 @@
 
 nostr:
   relay_url: "wss://relay.divine.video"
-  event_silence_timeout_secs: 300  # Resubscribe after 5 quiet minutes; fail health after 5 more
+  # Quiet window the listener tolerates before acting on relay silence.
+  # A cold silence resubscribes once, then fails health if a second window
+  # passes with no event. A relay CLOSED resubscribes immediately, so there
+  # the next quiet window is the one that fails health.
+  event_silence_timeout_secs: 300
   profile_relays:
     - wss://relay.divine.video
     - wss://purplepag.es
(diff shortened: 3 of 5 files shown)


NotThatKindOfDrLiz (approved):
Takeover fixes are on abb97fc. The recovery state now tolerates a queued sibling CLOSED, backs off after rate limiting, resets after healthy forwarding, avoids treating downstream backpressure as relay silence, validates the configured relay in subscribe output, and honors shutdown cancellation. The new regression coverage exercises each path. Local format, clippy, full tests, integration tests, and doc tests pass; GitHub test, semantic-title, and CLA checks are green.

realmeylisdev (approved):
Reviewed the whole branch including takeover commit `abb97fc`, and checked the recovery mechanism against the `nostr-relay-pool` 0.44 source rather than taking it on trust.

## Pushed (`abb97fc..84140bb`) — one commit per finding, each revertible alone

**`caacc56` — the repeated-closure test raced its own timeout.**
`a_second_owned_closed_message_fails_instead_of_replaying_again` used a `rate-limited:` closure reason, which routes the first recovery through the real five-second `RATE_LIMIT_RECOVERY_BACKOFF` sleep. The test wraps `run_live_loop` in `timeout(PATIENCE)`, and `PATIENCE` is also five seconds — I measured it at exactly 5.00s on three consecutive runs. It passed only because `Timeout` polls its inner future before its own deadline and the work after the sleep happened not to yield; one extra scheduler round trip on a loaded runner turns it into "live loop did not react to repeated CLOSED". The rate-limit prefix was incidental to what the test asserts, and `rate_limited_closure_backs_off_before_recovery` already covers the backoff on a paused clock. Runtime 5.00s → 0.01s, with the full five seconds now genuine headroom.

**`0a8f6a4` — the config comment described only one of the two paths.**
Both shipped configs summarised the knob as "Resubscribe after 5 quiet minutes; fail health after 5 more". `recovery_attempted` is shared between the two recovery paths: the `CLOSED` handler sets it when it resubscribes immediately, and it clears only once an event reaches the handler queue. So after any owned-subscription closure — a routine relay restart — the first quiet window takes the `if recovery_attempted` branch and fails the listener outright. The budget there is one quiet window and zero further resubscribes, not "5 then 5 more". The behaviour is deliberate, since the closure already spent the resubscribe; only the description was wrong, and it is the line an operator tuning this number reads. Reworded in both configs and the README row. No behaviour change.

**`84140bb` — added the missing regression test for the watchdog rearming.**
Nothing walked silence → flow → silence. `silent_subscription_resubscribes_then_fails_listener_health` only walks silence → silence, where the listener is *meant* to exit, and `healthy_flow_allows_a_later_closed_recovery` walks CLOSED → flow → CLOSED. So nothing asserted that resumed event flow clears `recovery_attempted` on the silence path — the property that keeps an ordinary quiet spell from compounding into a restart. I checked the test earns its place by mutation rather than by passing: bundling the reset into the `closed_recovery` arm, a plausible refactor slip, leaves every other test in the module green and fails only this one.

## One thing worth your call before merge — recovery has no upper bound

A relay that keeps closing the subscription while still trickling events through drives resubscription indefinitely, with no growth in the backoff and no health failure.

Each cycle runs: `CLOSED` → recover → one event arrives, clearing `recovery_attempted` and setting `flow_resumed` → once `CLOSED_RECOVERY_GRACE` passes, `healthy_grace_elapsed` clears `closed_recovery` → the next `CLOSED` is treated as brand new. I drove ten such cycles against `run_live_loop`: ten recoveries, listener still alive, `Ok(())`. Sustained, that is one REQ replaying a one-hour `since` window roughly every ten seconds (`RATE_LIMIT_RECOVERY_BACKOFF` + `CLOSED_RECOVERY_GRACE`), indefinitely, against a relay that is plausibly closing us *because* we are over a limit.

Two reasons I am flagging this rather than fixing it:

- It runs against our relay-connection-hygiene guidance, which asks for exponential backoff capped around 60s and warns that reconnecting immediately is how you get rate-limited or banned. The five-second backoff here never grows.
- The end state is a pod that resubscribes forever, delivers a trickle, and reports healthy — the same shape as the pod in #56.

The remedy is a tuning decision rather than a mechanical fix — grow the backoff across consecutive recoveries, cap recoveries per window, or fail health after N — so it is yours to make. I do not think it blocks merge: it needs a relay that both closes us repeatedly and keeps delivering, which is not what #56 describes.

## Verified rather than assumed

- `RelayMessage::Closed` does reach the notification stream. `handle_relay_message` always ends with `send_notification(RelayNotification::Message …)` after its `Remove`/`MarkAsClosed` bookkeeping, so the immediate-recovery path is real.
- Recovery repairs both SDK outcomes. `Remove` drops the subscription, and `subscribe_with_id` → `update_subscription`'s `entry(id).or_default()` recreates it with the right defaults. `MarkAsClosed` leaves `closed: true` sticky, which is harmless — `should_resubscribe` reads it as *should* subscribe.
- The main-relay `output.success` check cannot false-negative. `AppState` puts only `nostr.relay_url` on `nostr_client` (`profile_relays` live on a separate client), and `RelayUrl` equality normalises trailing slashes. It also does not misfire on a transient disconnect: `ensure_operational` only returns `NotConnected` after repeated failures at a low success rate, so a routine reconnect still queues the REQ.
- A sibling closure arriving *after* `CLOSED_RECOVERY_GRACE` with no event in between does return `Err` and restart. That needs the relay to be closing subscriptions *and* delivering nothing, where a restart is the outcome you want, so I left it alone.
- The one-hour replay is safe against duplicate pushes: `processed_event_ttl_secs` is 604800.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all-features` — 149 lib tests plus every integration target, all green at `84140bb`. The branch is two commits behind `main`; the only overlap is one unrelated README line, so no conflict.

Approving.

@dcadenas the head moved to `84140bb` — please look over the three commits before you merge. The unbounded-recovery point above is the one thing I would want your decision on.

@NotThatKindOfDrLiz your approval was on `abb97fc`, so my push made it stale — re-requesting review.

PR #60, opened 2026-08-24
docs(push): explain when device registrations are removed

### LINKED ISSUE #55: chore(spec): the service ignores the expiration tag its own draft spec requires
Split out of #43, which had grown two unrelated halves. The dead-token half was fixed in #47; the "live tokens expire 90 days after registration" half is fixed in #54. What is left is a decision, not a defect, so it gets its own issue.

## The decision

Does this service honour the NIP-40 `expiration` tag on kinds 3079/3080, or does our draft spec drop the requirement?

Right now it does neither, and that is the problem.

## What's actually true today

`docs/nip-xx-push-notifications.md` is unambiguous about the requirement:

- Line 58: ``- `p`, `app`, `expiration` MUST be present.``
- Line 59: `- Expiration per NIP-40. Servers MUST ignore expired events. Clients SHOULD refresh early (30–90d).`
- Line 183: lists expiration under Security as what bounds replays.

The service does not implement any of it. `expiration` appears nowhere in `src/` — the only matches in the repo are in that doc and an unrelated TTL comment in `src/preferences.rs`. The registration handler never reads the tag, never validates it is present, and never rejects an expired event.

Meanwhile both sides independently picked 90 days:

| | Value | Source |
|---|---|---|
| Client `expiration` | now + 90 days | `divine-mobile` `mobile/lib/services/push_notification_service.dart:46` — `static const pushTokenExpirationDays = 90` |
| Server sweep | 90 days, inclusive | `config/settings.yaml` — `token_max_age_days: 90`, applied by `cleanup_stale_tokens` in `src/redis_store.rs` |

Those two numbers agree by coincidence, not by design. Nothing links them. Either one can be changed without the other noticing, and the drift would be silent — no error, no warning, no test failure.

### The requirement is already dead letter on the client too

The 90-day value on the client matches the draft's own JS example exactly (`+ 7776000` at line 205 of the doc — 90 x 86400), so 3079 does follow the spec. But the tag stops there.

`expiration` appears exactly once in `divine-mobile`'s `push_notification_service.dart`, at line 366, inside the kind 3079 builder. Kind 3080 gets only `p` and `app`:

```dart
// mobile/lib/services/push_notification_service.dart:237-240
List<List<String>> _deregistrationTags(String pushServicePubkey) => [
  ['p', pushServicePubkey],
  ['app', pushAppIdentifier],
];
```

Our draft says of 3080: *"Same structure and rules as 3079."* — and its example for 3080 shows an `expiration` tag. So the client omits a tag the draft marks MUST. Kind 3083 omits it too, though the draft does not require it there.

That matters for the decision below. The `expiration` requirement is not a rule we implement imperfectly; it is a rule nothing on either side honours. The server never reads it on any kind, and the client sends it on one kind out of three. Option 1 is therefore new work everywhere, not a gap-fill — which is worth weighing against the fact that nothing has depended on it so far.

## Why it needs deciding rather than just implementing

The obvious move — "implement what the draft says" — is not obviously right.

Honouring `expiration` literally means dropping a registration once its tag passes, on a timer, with no reference to whether the device is actually reachable. That is the same failure mode #54 just removed from the sweep: time is a poor proxy for whether a device exists.

The nearest thing to a standard here, [nostr-protocol/nips#1528](https://github.com/nostr-protocol/nips/pull/1528), deliberately went the other way. It relies on provider feedback rather than synchronising each token's lifecycle against a clock. Since #47 the provider feedback path works — `UNREGISTERED` and `SENDER_ID_MISMATCH` now reach the classifier and prune on positive evidence — so we have the mechanism that PR prefers.

There is also a replay argument on the other side: the draft cites expiration as what bounds replay of a registration event. Dropping the requirement means finding another answer for that, or accepting the exposure. That is worth a deliberate call rather than a silent omission.

## Options

1. **Implement it.** Read and validate `expiration` on 3079/3080, reject expired events, and derive the sweep window from the tag instead of a separate config value so the two cannot drift.
2. **Drop it from the draft.** Remove the MUST from `docs/nip-xx-push-notifications.md`, follow nostr-protocol/nips#1528 in relying on provider feedback, and give the replay concern its own answer.
3. **Something in between** — for example require the tag and validate it, but keep provider feedback as the authority on deletion.

Whichever way it goes, the doc and the code should stop disagreeing.

## Not covered here

The residual gap in #43 that neither #47 nor #54 closes: a user who receives no notifications at all for 90 days is still swept, because a delivery-based refresh cannot help someone nothing is delivered to. That is a consequence of this same decision and should be settled with it.

## Provenance

Server-side facts verified by reading the repo at `7f23522`. The client-side constant and the absence of a persisted re-registration gate were verified by reading `divine-mobile` at `e358fbfb68`: registration is driven by `handleReadiness` in `mobile/lib/services/push_notification_session_coordinator.dart:84`, gated only by in-memory state, so a 3079 is republished on every app start that reaches a ready session. `divine-mobile` declares no background task package (no `workmanager`, no `background_fetch`), so a dormant install cannot republish on its own. `UIBackgroundModes` does declare `fetch`, but no fetch handler is implemented anywhere in `ios/`, so it cannot drive a re-registration; Android has no WorkManager, JobService, BOOT_COMPLETED or AlarmManager path, and no `onNewToken` override. There is also no client-side token rotation or staleness handling at all — nothing on the client reads or acts on the `expiration` value it writes.

NotThatKindOfDrLiz:
## Summary

The draft push-notification specification said every registration and deregistration needed an expiration timestamp and that the service rejected expired events. The service has never interpreted that tag, so the published contract disagreed with both server and client behavior.

This change makes the expiration tag optional for relay interoperability and documents the lifecycle the service actually implements: explicit deregistration, terminal provider feedback, or the configured inactivity sweep. It also describes the existing replay controls without presenting the event expiration timestamp as one of them.

## Why this is the right fix

A registration timestamp is not evidence that a device has stopped working. The service already removes tokens on positive provider feedback and, since #54, refreshes token activity after successful delivery. Treating an event expiration timestamp as a separate deletion clock would restore the same time-based failure mode that activity tracking was introduced to avoid.

## What this deliberately leaves alone

- No server or client behavior changes.
- The service continues to accept an optional NIP-40 expiration tag without interpreting it.
- A token that receives neither a registration event nor a successful delivery during `token_max_age_days` can still be removed by the inactivity sweep; changing that policy is outside this issue.
- Future-dated event handling is unchanged.

## Verification

- `cargo fmt --check`
- `cargo test` — 190 tests and 1 doc test passed
- Checked every remaining expiration reference for consistency

Closes #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The service implements a draft push-notification protocol; see [docs/nip-xx-push
 - **User preferences** — users can opt in or out of notification kinds with a preferences event; sensible defaults apply otherwise.
 - **Deduplication** — atomic Redis `SET NX EX` per-event locks ensure each event is delivered once, even across replicas.
 - **Replay protection** — a configurable processing window (7 days by default) ignores stale events.
-- **Token cleanup** — a background task prunes stale tokens (older than 90 days by default) once a day.
+- **Token cleanup** — a daily background task prunes tokens with no registration or successful delivery for 90 days by default.
 - **Prometheus metrics** — relay ingestion, event processing, FCM delivery outcomes, and token pruning are exposed for monitoring.
 - **Optional allow-list** — `allowed_pubkeys` can restrict delivery to a specific set of recipients.
 
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -386,6 +386,9 @@ the fan-out follow-up rather than done by halves here.
 
 ## Redis Keys
 
+The canonical registration and removal rules live in the push specification's
+[Token lifecycle](nip-xx-push-notifications.md#token-lifecycle) section.
+
 | Key Pattern | Type | Description |
 |-------------|------|-------------|
 | `user_tokens:{pubkey}` | Set | FCM tokens registered for a pubkey |
--- a/docs/nip-xx-push-notifications.md
+++ b/docs/nip-xx-push-notifications.md
@@ -33,6 +33,9 @@ read it. It is not one of the control kinds above and carries no ciphertext.
 
 ### Registration (kind 3079)
 
+The `expiration` tag in this example is optional. Clients may include it for
+NIP-40 relay interoperability, but this service does not interpret it.
+
 ```json
 {
   "kind": 3079,
@@ -55,13 +58,14 @@ The content field contains the NIP-44 encrypted token payload. Example plaintext
 **Note:** The exact payload structure is implementation-specific. Services define their own required fields.
 
 **Rules:**
-- `p`, `app`, `expiration` MUST be present.
+- `p` and `app` MUST be present. `expiration` MAY be present.
 - `content` MUST be NIP-44 ciphertext; services MUST reject plaintext.
-- Expiration per NIP-40. Servers MUST ignore expired events. Clients SHOULD refresh early (30–90d).
+- This service does not use `expiration` to accept, reject, or remove a token.
 
 ### Deregistration (kind 3080)
 
-Same structure and rules as 3079.
+Uses the same encrypted token payload and required `p` and `app` tags as 3079.
+It MAY also carry an `expiration` tag under the same optional policy.
 
 ```json
 {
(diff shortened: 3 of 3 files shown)


dcadenas (approved):
The expiration contract now matches the service's implemented token lifecycle and replay behavior.

### Changes I pushed

- Matched provider-pruning wording to the errors that actually remove tokens.
- Defined inactivity using both accepted registrations and successful deliveries.
- Aligned the README cleanup summary with the canonical lifecycle section.

### Checks

- `cargo test --verbose`
- Gap: the reusable Semantic PR workflow is not runnable locally; its GitHub check is green.
- The GitHub `test` check was still running when this approval was submitted.

PR #61, opened 2026-08-24
fix(relay): restart listeners stuck in repeated subscription recovery

NotThatKindOfDrLiz:
## Summary

A relay that repeatedly closes subscriptions while still sending occasional events can keep a listener healthy and trigger one-hour replay requests indefinitely. This follow-up to #57 bounds that recovery churn so Kubernetes can replace a listener that never stabilizes.

Related to #56.

## Why this fix

The listener now tracks relay-driven recoveries in a rolling window equal to the configured event-silence timeout. It allows three recoveries in that window, then fails listener health before issuing a fourth replay. Isolated relay restarts age out of the window and keep the existing recovery behavior.

This does not change silence-triggered recovery, the one-hour replay horizon, or the existing five-second rate-limit delay.

## Verification

- `cargo test --all-features`

Regression coverage drives repeated `CLOSED` → rate-limit backoff → trickle event cycles and verifies that the fourth closure fails health without issuing another replay.

### DIFF
--- a/src/nostr_listener.rs
+++ b/src/nostr_listener.rs
@@ -43,6 +43,11 @@ const CLOSED_RECOVERY_GRACE: Duration = Duration::from_secs(5);
 /// Limit relay-driven recovery churn within one silence-watchdog window. A
 /// relay that repeatedly closes subscriptions while still trickling events can
 /// otherwise keep the listener healthy and replay the lookback indefinitely.
+///
+/// The window is the configured event-silence timeout, not a separate knob, so
+/// raising `event_silence_timeout_secs` also widens the span these recoveries
+/// are counted over and makes the budget easier to trip. Keep that in mind when
+/// tuning the silence timeout.
 const MAX_CLOSED_RECOVERIES_PER_WINDOW: usize = 3;
 
 /// How long to wait for the initial relay connection before treating startup as
--- a/src/nostr_listener.rs
+++ b/src/nostr_listener.rs
@@ -10,7 +10,7 @@ use crate::{
 };
 use async_trait::async_trait;
 use nostr_sdk::prelude::*;
-use std::collections::HashSet;
+use std::collections::{HashSet, VecDeque};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


mbradley (approved):
Approving. I checked this against the original bug report (#56) and the earlier fix that followed it (#57). The bug report, this PR's description, and the code all line up.

**The issue (#56):** a pod could stay marked healthy while receiving almost no events, with no way to detect it — one pod ran that way for 42 hours. #57 taught the listener to notice a dropped subscription and re-subscribe.

**What this PR adds:** #57 left one hole. If a relay keeps dropping the subscription but still sends a few events, the pod keeps looking healthy and keeps re-requesting an hour of history, over and over, with no limit. This change counts those recoveries: it allows three within one window, then fails the health check so the pod is replaced instead of recovering forever. The description says exactly this, and the code does exactly that.

**For us running the service:** a problem that used to be invisible now shows up as a pod restart. The tradeoff: against a relay that stays broken for a long stretch, rather than just briefly, the pod will restart repeatedly instead of sitting quietly. That is louder on purpose, and the restart rate is capped by Kubernetes backoff.

**For our users:** no change when things work. When a pod is stuck, it now gets replaced quickly, so people are more likely to actually receive their notifications. The one-hour replay after a restart, plus the existing de-duplication, means no one is notified twice for the same event.

**Two small commits I added to the branch:**
- A test. The description promises that relay restarts spaced far apart keep recovering normally, but nothing tested it — every existing test packed the recoveries close together. The new test spaces them out and checks the pod keeps recovering. I confirmed it catches a real regression by breaking the aging logic and watching the test fail.
- A one-line comment. The window used to count recoveries is the same setting as the event-silence timeout, so raising that timeout also makes this limit easier to hit. That link wasn't written down anywhere, so I noted it at the setting.

Full test suite, linter, and formatting all pass.

dcadenas (approved):
✅ repeated CLOSED recovery now terminates before a fourth replay.

PR #62, opened 2026-08-24
fix(deploy): complete intentional production rollouts

### LINKED ISSUE #59: Fix manual production deploy dispatch authorization
## Problem

The `Build, Test & Push` workflow can publish an image to the production registry through `workflow_dispatch` with `include_production: true`, but its subsequent `image-deploy` repository dispatch does not include any production authorization field.

`divine-iac-coreconfig` correctly rejects production dispatches unless the payload contains one of `hotfix: true`, `internal: true`, or `auto_promote: true`. As a result, the source workflow reports success after publishing the production image, while the downstream production rollout fails closed during payload validation.

This occurred while deploying commit `3d64b9f6ab1b92038b25d5d70139e077dd096648` from PR #54. The rollout required a separate production-only dispatch with `hotfix: true`.

## Expected behavior

The supported manual production deployment path should send a coreconfig payload that satisfies the production authorization policy and accurately represents the selected release mode. It must not weaken or bypass coreconfig validation.

## Acceptance criteria

- Define the intended production mode for manual `include_production` deployments (`hotfix`, internal promotion, or automatic promotion).
- Include the corresponding authorization field in the `image-deploy` client payload.
- Keep ordinary pushes to `main` scoped to POC and staging.
- Add workflow-level validation or tests that pin the production payload contract.
- Verify a manual production run publishes the image, creates the production coreconfig deployment PR, and reaches the ArgoCD sync without requiring a second dispatch.
- Update deployment documentation if the operator-facing procedure changes.

## Evidence

- Source production workflow run: https://github.com/divinevideo/divine-push-service/actions/runs/32676819532
- Rejected downstream dispatch: https://github.com/divinevideo/divine-iac-coreconfig/actions/runs/32676929500
- Corrected production deployment PR: https://github.com/divinevideo/divine-iac-coreconfig/pull/1914

NotThatKindOfDrLiz:
## Problem

Manual production promotions and version-tag releases publish the production image, but their downstream deployment dispatch omits the authorization required by coreconfig. The source workflow therefore reports success while the production rollout fails validation and requires a misleading emergency-hotfix dispatch.

## Why this fix

The workflow now builds one tested dispatch payload for every release path. Version tags and manual runs with `include_production` include `production` and `auto_promote: true`, accurately recording an intentional promotion without adding a hotfix label. Main-branch and ordinary manual runs remain limited to POC and staging. The production push, workflow summary, and downstream dispatch all consume this same payload decision, and a fail-closed guard rejects any future production payload missing authorization.

This deliberately does not change coreconfig policy, add a production GitHub Environment, or wait for the downstream workflow conclusion. Emergency `hotfix: true` dispatches remain an out-of-band platform operation.

## Verification

- `scripts/deploy-payload-contract-test.sh`
- `bash -n scripts/build-deploy-payload.sh scripts/deploy-payload-contract-test.sh`
- `shellcheck scripts/build-deploy-payload.sh scripts/deploy-payload-contract-test.sh`
- `actionlint .github/workflows/publish-and-release.yml`
- `cargo test --verbose`
- Negative guard check: a production payload with `auto_promote` removed is rejected

## Rollout

After merge, run the workflow manually with `include_production` selected and verify the downstream coreconfig deployment PR merges without a `hotfix` label and the ArgoCD sync succeeds.

Closes #59

### DIFF
--- a/.github/workflows/publish-and-release.yml
+++ b/.github/workflows/publish-and-release.yml
@@ -1,5 +1,6 @@
-# Builds once, pushes to all GCP Artifact Registry environments via docker-push-gcp.
-# Production push only on version tags (v*). Dispatch to IAC repo triggers ArgoCD sync.
+# Builds once, pushes to GCP Artifact Registry via docker-push-gcp, and dispatches
+# the matching environments to the IAC repo. Version tags and opted-in manual runs
+# include production as an intentional automatic promotion.
 #
 # Required GitHub Variables (repo-level):
 #   Per environment: WORKLOAD_IDENTITY_PROVIDER_{ENV}, SERVICE_ACCOUNT_{ENV}, GCP_PROJECT_ID_{ENV}
@@ -53,6 +54,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Test deploy payload contract
+        run: scripts/deploy-payload-contract-test.sh
+
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
@@ -82,6 +86,8 @@ jobs:
       packages: write
     outputs:
       tags: ${{ steps.tags.outputs.tags }}
+      deploy_payload: ${{ steps.deploy-payload.outputs.payload }}
+      include_production: ${{ steps.deploy-payload.outputs.include_production }}
 
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
Authorized production dispatches now fail closed without exposing tag-derived data to the shell.

### Changes I pushed

- Moved ref-derived tags out of the shell script.
- Made payload parsing fail closed and pinned version-tag manual runs.
- Removed an unsupported runbook claim.

### Checks

- Declared Cargo test, Clippy, and formatting commands passed.
- Payload contract: 5 cases passed; Bash syntax passed.
- Gaps: the declared-matrix tool cannot run the repo-local contract command or reusable semantic-title workflow. The contract passed manually and the live semantic check passed. `shellcheck` and `actionlint` were unavailable locally.

mbradley (commented):
What I understand exist(s/ed): when we deploy this service, the workflow here builds the image and hands off to `divine-iac-coreconfig`, which opens and merges a deploy PR that ArgoCD then syncs to the clusters. Coreconfig refuses a production deploy unless the handoff carries one of three "yes, production" flags. Today this workflow sends none of them, so production handoffs are rejected — while the workflow still reports success. The last production release had to go through as a manual emergency request to get around this.

This PR sends `auto_promote: true` on the production paths (version tags, and manual runs that tick `include_production`), which coreconfig accepts — so production deploys now complete cleanly without the emergency workaround. It also builds the whole handoff in one place (removing a duplicated condition that could drift) and adds a guard that refuses to send a production deploy missing that flag. The fix is correct: coreconfig accepts `auto_promote`, the contract test passes, and the payload clears coreconfig's full validation.

This means production flows differently after this: normal merge to `main` deploys to staging only. Production happens only when someone cuts a `v*` tag or manually runs the workflow with `include_production`. That deliberate trigger is the decision point — after it, coreconfig's deploy PR is merged automatically (that's how every service here deploys; the deploy PR isn't human-reviewed) and ArgoCD syncs. So merging this PR doesn't deploy to production; it makes the deliberate production trigger actually work.

This aligns with how funnelcake is deployed: it uses the same `auto_promote` flag for deliberate production dispatches, keeping `hotfix` for emergencies — so the flag choice here matches precedent. No service requires a second person to approve a production deploy; the deliberate trigger is treated as the decision.

One question, not a blocker: funnelcake puts production in its own workflow with a required "reason" field (a per-deploy audit trail) and a named `production` environment. This service uses a checkbox on the build workflow, with neither. Both patterns are in use. The question for the owners: should push-service adopt funnelcake's more auditable production path, or is the lighter checkbox fine here? Either way the code in this PR is correct.

Leaving this as a comment rather than an approval: @dcadenas owns this deploy machinery (coreconfig/ArgoCD), so I'd defer the approval and the ceremony question above to him.

Issue #65, opened 2026-08-27
feat(push): deliver push for NIP-17 direct messages (kind 1059)


NotThatKindOfDrLiz:
## Why

DM push is a P1 launch requirement, and there is no path to it today. A user who
receives a NIP-17 direct message — including a reply from Divine Moderation —
gets no notification at all. They find out by reopening the app and looking.

The service subscribes to a configured `event_kinds` list (`src/config.rs:242-259`)
and gates delivery on the user's kind-3083 preferences. Kind 1059 appears nowhere
in the source. Mobile's preference model is the same shape and equally silent:
`toKindsList()` emits `7, 1, 3, 16, 34236` (`divine-mobile mobile/lib/models/notification_preferences.dart:76-83`).

## Correcting the reason this was thought to be blocked

`divinevideo/support-trust-safety#192` records that gift-wrapping hides the
recipient, so the push service cannot know who to notify. That is backwards.
Per NIP-59, a gift wrap carries the recipient in the clear for routing:

```json
{ "kind": 1059,
  "pubkey": "<random, one-time-use pubkey>",
  "tags": [["p", "<recipient pubkey>"]] }
```

The wrap hides the *sender* and the content. The spec is explicit: "A gift wrap
can add metadata (recipient, tags, a different author) without revealing the true
author." It is the seal (kind 13) that carries no `p` tag. The same issue's
Finding 2 depends on this being true — it is how "who messaged support, and when"
is publicly enumerable today.

So a `#p`-filtered 1059 subscription is mechanically the same shape this service
already runs for kinds 7, 1 and 16.

## The decision to make before writing the watcher

NIP-59 says relays SHOULD gate kind 1059 behind NIP-42 and serve wraps only to
the marked recipient. `support-trust-safety#192` Finding 2 asks for exactly that
hardening on `wss://relay.divine.video`, which currently reports
`limitation.auth_required = false`.

A push watcher reading every registered user's 1059s needs the opposite: a
privileged cross-recipient read. Those two directions have to be reconciled up
front.

- Build the watcher against the open relay now, and it goes dark the day the
  relay hardens — silently, since an authenticated-empty result is
  indistinguishable from no traffic.
- Harden first, and the watcher needs a designed privileged-reader path from the
  start (service identity with NIP-42 AUTH and a relay-side allowance, or an
  internal delivery hook that does not read the public relay at all).

This issue is asking for that call, then the implementation that follows it.

## Shape of the work, once the direction is chosen

[...]

PR #66, opened 2026-08-28
feat(push): notify users about moderation messages

dcadenas:
## Summary

Moderation direct messages are encrypted, so the push service cannot tell a real incoming message from a sender copy, reaction, or file by watching gift wraps on public relays.
This adds a small authenticated internal hook that accepts an already-classified moderation message and sends a generic, content-free push to the recipient.

## What Changed

The service now accepts bearer-authenticated requests for automated moderation notices and report outcomes.
It reuses the existing preference, token, deduplication, pruning, and Firebase delivery paths without giving the push service a NIP-17 decryption key or adding kind 1059 to its relay subscription.

- Authenticates before parsing the request body.
- Uses the gift-wrap event ID and recipient for replay deduplication.
- Sends only generic `New message` copy and omits sender, content, timestamp, and reference metadata.
- Documents the endpoint's best-effort response semantics and rollout requirements.

The moderation-service caller is in companion draft PR divinevideo/divine-moderation-service#210.

## Testing

- `cargo test --verbose`
- `scripts/deploy-payload-contract-test.sh` (5/5 cases)

The semantic PR workflow has no local command and will run in GitHub.

## Risks

Do not enable callers until Divine Mobile publishes kind 1059 preferences, bumps its published-kinds schema, and routes `directMessage` taps to the message inbox.
A reachable internal URL and shared bearer secret also need platform provisioning.

A `204` means the request was processed, not that a device received the push.
The caller is deliberately best-effort and does not retry, so push failures never delay or fail moderation.
Manual moderator replies and community warnings are outside this initial two-class contract.

Refs #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Any setting can be overridden with the `NOSTR_PUSH__` prefix and `__` as the nes
 | `NOSTR_PUSH__REDIS__URL` | No | Redis connection URL (overrides the config file) |
 | `NOSTR_PUSH__NOSTR__RELAY_URL` | No | Nostr relay to subscribe to |
 | `NOSTR_PUSH__NOSTR__EVENT_SILENCE_TIMEOUT_SECS` | No | Quiet period before the listener resubscribes, and the window after any resubscribe before it fails health (default `300`) |
+| `NOSTR_PUSH__SERVER__INTERNAL_API_TOKEN` | No | Shared bearer token that enables authenticated internal push requests |
 | `APP_ENV` | No | Selects the config file (default `development`) |
 | `RUST_LOG` | No | Log level (default `info`) |
 
--- a/config/settings.development.yaml
+++ b/config/settings.development.yaml
@@ -59,8 +59,10 @@ notification:
       - 1      # Text notes (comments, mentions)
       - 7      # Reactions/likes
       - 16     # Reposts
+      - 1059   # Direct messages from classified internal hooks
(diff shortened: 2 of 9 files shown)


NotThatKindOfDrLiz:
I pushed `2914146` to finish the review fixes:

- added production-path coverage for runtime token wiring, the disabled endpoint, DM preference/no-token gates, recipient deduplication, and retryable claim release;
- removed the unreachable DM lazy-copy state and restored the unrelated payload-order/comment churn;
- sanitized internal error responses and mapped cancellation/retryable delivery to `503`;
- added bounded internal-request outcome metrics and case-insensitive bearer-scheme parsing;
- documented the endpoint's 45-second FCM ceiling and the caller-side requirement to keep it off the moderation critical path.

Local verification passed: `cargo fmt --check`, `cargo check`, `cargo clippy --all-targets --all-features -- -D warnings`, and the complete `cargo test` suite with Redis-backed tests running (186 unit tests plus all integration/doc tests).

@dcadenas, please review the takeover changes. Once CI and the required human review are green, this should be ready for you to merge.

NotThatKindOfDrLiz (approved):
Reviewed the full diff at 2914146 after the takeover fixes. Runtime token wiring, fail-closed routing, real direct-message delivery gates, deduplication, retryable claim release, sanitized errors, bounded auth metrics, and the caller latency contract are now covered. Local verification and required CI are green.

